PR #1, opened 2018-01-24
Enable smartypants for curly quotes, ... ellipsis, -- en dash, --- em…

tmcw:
… dash.

Fixes https://github.com/observablehq/notebook-runtime/issues/73 - saves us seconds or perhaps hours of meticulously curling quotes.

### DIFF
--- a/src/md.js
+++ b/src/md.js
@@ -5,7 +5,7 @@ export default function(require) {
         var string = strings[0] + "", i = 0, n = arguments.length;
         while (++i < n) string += arguments[i] + "" + strings[i];
         var root = document.createElement("div");
-        root.innerHTML = marked(string, {langPrefix: ""}).trim();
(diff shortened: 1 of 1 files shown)


tmcw:
Our options here are:

- Keep the current behavior
- Invest some time on getting `marked`'s smartypants mode up to par with Medium's intelligence and then turn smartypants on
- Turn smartypants on, and set a backlog item to improve the quote-curling

My preference is for the last one, of course, because smartypants is only wrong a fraction of the time and in those cases we can override it by adding an explicit-direction-quote, and also because even of the really well-written notebooks out there, most people who aren't us aren't going to spend the time typing curled quotes and `...` etc when most of the time platforms do that for them.

tmcw:
Closing; still in favor of enabling smartypants, but we'd need consensus to do it.

PR #20, opened 2018-03-19
Update d3-require.

mbostock:
image


mbostock:
This also works:

image

PR #28, opened 2018-04-13
Add Generators.{disposable,worker}.

mbostock:
These methods make it easier to clean up resources that need explicit disposal, such as workers and GPU-backed tensors. For example, a dedicated web worker:

```js
worker = Generators.worker(`
onmessage = function ({data}) {
  postMessage({echo: data});
};
`)
```

A tensor:

```js
x = Generators.disposable(tf.tensor2d([[0.0, 2.0], [4.0, 6.0]]), x => x.dispose())
```

### DIFF
--- a/src/generators/dispose.js
+++ b/src/generators/dispose.js
@@ -1,8 +1,8 @@
-export default function finalize(value, finalizer) {
+export default function dispose(value, disposer) {
   let done = false;
   return {
     next: () => done ? {done: true} : (done = true, {done: false, value}),
-    return: x => (done = true, finalizer(value), x),
+    return: x => (done = true, disposer(value), x),
     throw: () => ({done: done = true})
   };
 }
--- a/src/generators/worker.js
+++ b/src/generators/worker.js
@@ -1,9 +1,9 @@
-import dispose from "./dispose";
+import disposable from "./disposable";
 
 export default function worker(source) {
   const url = URL.createObjectURL(new Blob([source], {type: "text/javascript"}));
   const worker = new Worker(url);
-  return dispose(worker, () => {
+  return disposable(worker, () => {
     worker.terminate();
     URL.revokeObjectURL(url);
   });
--- a/src/generators/worker.js
+++ b/src/generators/worker.js
@@ -1,9 +1,9 @@
-import finalize from "./finalize";
+import dispose from "./dispose";
 
 export default function worker(source) {
   const url = URL.createObjectURL(new Blob([source], {type: "text/javascript"}));
   const worker = new Worker(url);
-  return finalize(worker, () => {
+  return dispose(worker, () => {
     worker.terminate();
     URL.revokeObjectURL(url);
   });
--- a/src/generators/worker.js
+++ b/src/generators/worker.js
@@ -1,0 +1,10 @@
+import finalize from "./finalize";
+
+export default function worker(source) {
+  const url = URL.createObjectURL(new Blob([source], {type: "text/javascript"}));
+  const worker = new Worker(url);
+  return finalize(worker, () => {
+    worker.terminate();
+    URL.revokeObjectURL(url);
+  });
+}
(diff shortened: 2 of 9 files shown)


src/generators/dispose.js
```diff
@@ -0,0 +1,8 @@
+export default function dispose(value, disposer) {
+  let done = false;
+  return {
+    next: () => done ? {done: true} : (done = true, {done: false, value}),
+    return: x => (done = true, disposer(value), x),
```
jashkenas: The triple comma is slightly cryptic, but I like it!

tmcw:
_Not entirely in the scope of this - most likely something for userspace but I'd like to get handle on whether it'd be possible to do this in userspace:_

What would an equivalent to `util.promisify` look like for 'disposable objects'? Like would this make it possible to do something like:

```js
const {tensor2d} = disposable(tf.tensor2d);

let a = tensor2d([[1, 2]]); // automatically disposable fancy version of tensor2d
```

src/generators/worker.js
```diff
@@ -0,0 +1,10 @@
+import dispose from "./dispose";
+
+export default function worker(source) {
+  const url = URL.createObjectURL(new Blob([source], {type: "text/javascript"}));
+  const worker = new Worker(url);
```
tmcw: the API for [SharedWorker](https://github.com/conventional-changelog/standard-version) is super similar - maybe we could support either by making this `worker(source, constructor)`?

nsthorat:
In general, I *love* the approach, this is going to really help!

 +1 I think "dispose" may not be the right name (it seems like it's going to dispose immediately). "disposable" seems reasonable, and is a pretty common pattern. See [closure](https://google.github.io/closure-library/api/goog.Disposable.html) as well as the C# link above. I think "finalize" would be reasonable too - in a sense Observable is destroying the previous value of the cell and giving you a callback before it happens.

So a typical cell using tfjs might look like this (correct me if I'm missing something):

```
{
  const y = tf.tidy(() => {
    const x = tf.scalar(2);
    return x.square();
  });
  return Generators.disposable(y, y.dispose());
}
```

One thing we might do as our own TF.js + observable stdlib notebook is something like this:
```
{
  return tflib.generateDisposable(() => {
    const x = tf.scalar(2);
    return x.square();
  });
}
```

Where `tflib` is imported from some TF.js stdlib notebook.

Obviously I need to think about the naming of "generateDisposable" in this case, but you get the drift (it tidy()'s and also disposes that last value that was generated).

You guys rock!

nsthorat:
I just saw @tmcw comment with `const {tensor2d} = disposable(tf.tensor2d);`.

The only issue with this is that usually the tensor you're returning is the output of some mathematical operation, which means we'd have to hijack all of our operations to make them return generators (see the snippet I just linked above).

I think generally speaking, the last code snippet I posted above is what a common cell will look like.

mbostock:
Thanks for the feedback, @nsthorat! I’ve renamed to Generators.disposable and I’m happy with that name.

You can shorten your cell definition slightly since the *dispose* function is passed the return value:

```js
y = Generators.disposable(tf.tidy(() => {
  const x = tf.scalar(2);
  return x.square();
}), y => y.dispose())
```

And yeah, TensorFlow.js could provide its own disposable helper around Generators.disposable and tf.tidy:

```js
tf.disposable = function(fn) {
  return Generators.disposable(tf.tidy(fn), t => t.dispose());
};
```

(This assumes that you always return a Tensor from your tidy function, which is probably a bad assumption, but you get the idea.) Then you could say:

```js
y = tf.disposable(() => {
  const x = tf.scalar(2);
  return x.square();
})
```

Which would be pretty great!

Also, if you wanted, you could extend this pattern to support async by passing in a tf *context* object to your tidy function. This context would expose the same constructors as the static *tf*, but anything allocated by that context would be automatically disposed when the promise resolves or rejects. That could look like this:

```js
y = tf.disposable(async tf => {
  const x = tf.scalar(2);
  await somePromise;
  return x.square();
})
```

Might be hard to remember you need the *tf* argument to their tidy function, though—if you forget it, then *tf* is the static TensorFlow namespace, and your code continues to work but leaks tensors. 😁

PR #29, opened 2018-04-13
Part of (compiler-refactor)

jashkenas:
Only a small change to the standard library:

* Renames `runtimeLibrary` to just plain `Library`, for consistency with **notebook-runtime**.

* Also exports the "static" namespaces from the standard library at the top level, so you can individually import DOM, Files, Generators or Promises if you choose to. (The notebook-runtime uses just Generators).

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ If you use NPM, `npm install @observablehq/notebook-stdlib`.
 
 ## API Reference
 
-<a href="#runtimeLibrary" name="runtimeLibrary">#</a> <b>runtimeLibrary</b>([<i>resolve</i>])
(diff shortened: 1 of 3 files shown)


mbostock:
I’m okay with renaming to Library and making it the default export. 👍 

However, I’m weary of also exporting the static namespaces. First, mixing default and named exports is a little awkward when transpiling back to ES5. But more importantly, it locks us in to those symbols being static, rather than being able to configure them dynamically when the library is constructed. And it feels weird that these namespaces are available statically, but the rest of stdlib (require, md, etc.) is not.

jashkenas:
Alright — changed to that the runtime has to go through a stdlib Library instance in order to access "observe" and "input".

Issue #27, opened 2018-04-08
Dependency versioning with require?


mbostock:
Top-level version pinning works fine:

```js
d3 = require("d3-selection@1")
```

But unfortunately Observable doesn’t currently observe the package.json for a require’d module, so if you load a module with dependencies, things can easily break:

```js
d3 = require("d3-geo@1")
```

Here d3-geo depends on d3-array, and d3-geo’s package.json says it depends on d3-array@1. But if the latest major release of d3-array is 2, then loading d3-geo@1 will load d3-array@2, and :boom:.

Here’s another example that breaks:

```js
d3 = require("d3-array@1", "d3-geo@1")
```

Now you get two identical copies of d3-array, because Observable doesn’t realize that d3-geo’s d3-array dependency is equivalent to the top-level require of d3-array@1.

This is especially a problem with d3-selection, because if you have multiple copies of d3-selection then d3.event ends up being broken. (A dependency will set d3.event on its internal copy of d3-selection, but the top-level d3-selection will have a different d3.event that is always null.)

So, probably what needs to happen is that d3-require should fetch the package.json before loading the module, so that it can do appropriate version resolution of dependencies. But this may prove to be complicated. Alternatively, we could have some top-level configuration like a lockfile that specifies how to resolve the versions of dependencies, but this could be cumbersome to maintain.

PR #42, opened 2018-05-10
Update d3-require.

mbostock:
This introduces the async require.resolve while deprecating resolve.

(I will update the package.json and yarn.lock when d3/d3-require#10 is merged and released.)


mbostock:
Fixes #27.

Issue #32, opened 2018-04-19
Generalize generator transforms?


mbostock:
We currently provide two utility methods for creating a new Generator from an iterator: [Generators.filter](https://github.com/observablehq/notebook-stdlib/blob/master/README.md#Generators_filter) and [Generators.map](https://github.com/observablehq/notebook-stdlib/blob/master/README.md#Generators_map). There are a few problems with these methods:

* If the input *iterator* is a Generator, the returned Generator does not wrap *generator*.return and *generator*.throw. Possibly, these methods should return an Iterator rather than a Generator, and be moved to Iterators.filter and Iterators.map. Or, these methods could “upgrade” an iterator to a generator, and wrap the return and throw methods as expected.

* If the input *iterator* is an async iterator, these methods don’t work (at all).

* If the input *iterator* iterates over promises, the generator doesn’t wait for the promise to resolve. Coming from Observable, we somewhat expect the *test* or *transform* functions to be passed resolved values rather than Promise objects. However, in the case of Generators.filter, it’s not clear how that should work, since the *test* function might return false for the last promise, but it’s too late because the generator has already yielded a promise!

It’s possible that the simplest thing to do would be to deprecate these methods along with the deprecated DOM methods #31. I haven’t found them especially useful, and given the desire to keep the standard library small, it might be good to prune them.

Issue #31, opened 2018-04-18
Deprecate some DOM methods.


mbostock:
Proposed: we deprecate the following methods in favor of the [html](https://github.com/observablehq/notebook-stdlib#html) or [svg](https://github.com/observablehq/notebook-stdlib#svg) tagged template literal.

* DOM.canvas
* DOM.element
* DOM.input
* DOM.range
* DOM.select
* DOM.svg
* DOM.text

None of these methods are especially helpful, so I think it’s pretty much always more idiomatic to use literal HTML. By deprecating, I mean these methods would continue to be supported in the standard library, but we would remove their documentation from the README, or at least relegate them to a deprecated section.

(That leaves DOM.context2d, DOM.download, DOM.uid, for those keeping score.)

tmcw:
👍 I think `DOM.svg` and `DOM.range` still get a lot of usage, even in new notebooks that we've been writing, but - sure, these are non-essential.

jashkenas:
I'm moderately fond of those methods because I have a general aversion to string interpolation where it can be avoided — so that `DOM.canvas(width, 500)` feels a little better than `html\`<canvas width="${width}" height="500"></canvas>\`` ... but I certainly see your point.

Yes, we should deprecate them by removing the documentation — but it seems like there's no harm in leaving the actual methods around.

mbostock:
You can shorten it slightly…

```js
html`<canvas width=${width} height=500>`
```

danburzo:
Could we salvage the `attributes` part of `DOM.element`, maybe as `DOM.attr`? 
It has a bit of logic (around namespaces) that would provide benefits over doing `setAttribute` / `setAttributeNS` manually on DOM elements that were previously created.

mbostock:
DOM.svg came up as an unnecessary impediment to adapting D3 examples to vanilla JavaScript. Other than require, DOM.svg is the only stdlib method used by most of the D3 examples. And it’s trivial to replace

```js
d3.select(DOM.svg(width, height))
```

with 

``js
d3.create("svg").attr("viewBox", [0, 0, width, height])
```

which is (sufficiently) equivalent.

This, to me, is strong evidence that we should make the standard library as small as possible.

mootari:
I believe DOM.element is still relevant, as it facilitates attribute composition and escaping. Some of the issues with template strings are outlined here: https://observablehq.com/@mootari/attribute-escaping-in-template-strings

[...]

PR #43, opened 2018-05-11
Notebook Downloads (Standard Library Changes)

jashkenas:
* Creates UMD and ES Module bundles of the standard library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.2",
   "license": "ISC",
   "main": "src/index",
+  "module": "dist/notebook-stdlib.js",
   "author": {
     "name": "Observable, Inc."
   },
@@ -29,6 +30,7 @@
     "jsdom": "^11.6.2",
     "rollup": "^0.52.0",
     "rollup-plugin-node-resolve": "^3.0.0",
+    "rollup-plugin-uglify": "^3.0.0",
     "tape": "^4.9.0",
     "tape-await": "0.1.0"
   }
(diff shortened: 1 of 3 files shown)


jashkenas:
@mbostock — if approved, I'd like to merge and publish this PR as version 1.0.0 on npm...

mbostock:
Per https://github.com/tc39/proposal-observable, `Observable` may be a thing in ECMAScript in the future, although it is just stage 1. I suppose we could use `observablehq` as the namespace?

mbostock:
My other concern #31 #32 is whether we should formally deprecate (though still support) the following methods:

* Generators.filter
* Generators.map
* DOM.element
* DOM.input
* DOM.range
* DOM.select
* DOM.text

That said I’d probably *not* deprecate:

* DOM.canvas
* DOM.svg

I find myself still using DOM.svg a lot because it sets the viewBox automatically, likewise DOM.context2d is useful for its automatic high DPI support, and might as well keep DOM.canvas for symmetry with DOM.svg and for WebGL.

In practice this would mean moving the documentation for the deprecated methods from README.md to a DEPRECATED.md, and perhaps adding these to the deprecated list for autocomplete.

jashkenas:
CSS and UMD namespaces updated across the board to `observablehq`. 

(Although, I still hold out hope that one day we'll be able to purchase observable.com)...

Just for expediency's sake: If we're not changing any of the functionality (yet), just the documentation, I don't think we need to make a decision on the deprecations before merging this branch and publishing a 1.0.0 — as long as the functionality remains in place, we can deprecate at any point afterwards.

mbostock:
Okay, that’s fair; we can punt on deprecation. 👍

PR #55, opened 2018-06-04
Make Generators.disposable iterable.

mbostock:
Related #40.


mbostock:
Enabling this type of usage might actually be bad, since yield* won’t call *generator*.return. Specifically, a yield* or for-of loop won’t call *generator*.return if the generator is already done (which is always is in the case of Generators.disposable); it only calls it if the loop breaks early.

PR #56, opened 2018-06-05
Add require.alias.

mbostock:
For example, to declare React’s missing UMD entry:

```js
react = require.alias({
  react: "react/umd/react.production.min.js"
})("react")
```

You can also use this to pin exact versions.


mbostock:
(We could also expose require.from while we’re at it…)

tmcw:
Would we be able to create default aliases? This is great and reduces the hackiness of my d3-require 
based implementation, but I think people will still simply run `require('react')` and it'll still be broken by default.

mbostock:
We could, but I’m not sure if it’s a good idea for us to wade into per-library defaults.

PR #57, opened 2018-06-09
Add optional initial_selection to DOM.select.

jaynel:


mbostock:
Thanks for the PR! We’re probably going to deprecate DOM.select and related methods (see #31) in the near future in favor of HTML tagged template literals, so I don’t think it makes sense to add new functionality at this time.

For a static dropdown, I’d say:

```js
html`<select>
  <option value="red">red</option>
  <option value="green" selected>green</option>
  <option value="blue">blue</option>
</select>`
```

For a data-driven one, something like:

```js
colors = ["red", "green", "blue"]
```
```js
html`<select>${colors.map(c => `
  <option value="${c}" ${c === "green" ? " selected" : ""}>${c}</option>`)}
</select>`
```

PR #72, opened 2018-11-09
Set xml:base attribute.

mbostock:
On broken browsers such as Firefox and Safari <12, this fixes same-page fragment identifiers (such as for text paths and gradients) without requiring DOM.uid.


mbostock:
I thought this fixed it but now I’m not so sure.

PR #79, opened 2018-12-07
Use body width.

mbostock:
This eliminates the hard-coded 28px, making the definition more generally useful.


mbostock:
No, the 28px was for the cell margins (14px on each size).

PR #80, opened 2018-12-07
Wrap with span, not div.

### LINKED ISSUE #73: html should use a SPAN wrapper, not DIV

mbostock:
Previously, there was an implicit DIV when using an html or md tagged template literal that returns more than one node. For example:

```js
html`Hello, <i>world</i>!`
```

This implicit DIV is weird if you’re embedding the HTML inside something else (rather than as a top-level cell value); it’s better to use a container that doesn’t affect the content layout, hence SPAN.

Fixes #73.

### DIFF
--- a/src/html.js
+++ b/src/html.js
@@ -5,5 +5,5 @@ export default template(function(string) {
   template.innerHTML = string.trim();
   return document.importNode(template.content, true);
 }, function() {
-  return document.createElement("div");
+  return document.createElement("span");
 });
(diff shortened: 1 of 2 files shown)


mbostock:
Are you saying that DIV with display: block is the right approach? Or is there another element with display: inline that we should use instead?

jashkenas:
I'm saying that technically, a `<div>` is the appropriate semantically meaningless element to wrap "flow content" like paragraphs ... but that in practice, since we don't know whether you intend to render inline content or block content, this change to a `<span>` is probably the best approach — despite not validating as HTML5.

PR #82, opened 2018-12-21
Lazy DOM.download.

### LINKED ISSUE #75: DOM.download, but lazy.
As shown here: https://beta.observablehq.com/@mbostock/lazy-download

mbostock:
Fixes #75.

* Allows the *value* to be specified optionally as a promise, a function, or a function that returns a promise. If a promise or a function that returns a promise, the first click changes the label to “Saving…” and disables the button while waiting for the promise to resolve, then changes the label to “Download”, and the second click will do the download.

* If the *value* is specified as a function, it is re-evaluated on each click.

* Changes the default label to “Save” instead of “Download” (for better compatibility with asynchronous values, where the “Saving…” and “Download” labels are applied internally and not customizable).

### DIFF
--- a/src/dom/download.js
+++ b/src/dom/download.js
@@ -1,11 +1,31 @@
-export default function(object, name, label) {
-  var a = document.createElement("a"),
-      b = a.appendChild(document.createElement("button"));
-  b.textContent = label == null ? "Download" : label;
-  a.download = name == null ? "untitled" : name;
-  a.onclick = function() {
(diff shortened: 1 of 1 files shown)


tmcw:
I’m okay with this but I do think that it’d be nice for the button to work with one click, if it detects that the value is _not_ a promise or function.

mbostock:
It does that already.

PR #87, opened 2019-03-13
Embed marked

tmcw:
One part of https://github.com/observablehq/notebook/issues/2197

Main gotcha that happens here is that the dependency path notebook - runtime - stdlib ends up with the 'module' entry in stdlib, and that means that it tries to import marked, and because marked is commonjs, it fails. So… we may need to either special-case that in the notebook, or something else.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
   "dependencies": {
     "d3-require": "^1.2.0",
     "esm": "^3.0.84",
-    "marked": "^0.6.1"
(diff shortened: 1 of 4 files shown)


tmcw:
In terms of doing this _the long way_, I opened a PR to add an ES Module entry point to marked https://github.com/markedjs/marked/pull/1452

mbostock:
Let’s fork marked and use that as a dependency for now.

mbostock:
Thanks. If you revert the rollup.config.js changes and remove the rollup-plugin-commonjs dep, this looks good. 👍

mootari:
Was the version jump from 0.3.12 to 0.6.1 intended?

PR #88, opened 2019-04-09
Polyglot highlight.js support

tmcw:
Downstream of https://github.com/observablehq/highlight.js/pull/2 and upstream of notebook-runtime, this adds the ability to highlight any of highlight.js's many highlightable languages without loading them all in off the bat.

### DIFF
--- a/src/md.js
+++ b/src/md.js
@@ -1,8 +1,8 @@
 import template from "./template";
 import marked from "marked";
 
-const LANGUAGE_ROOT =
-  "https://cdn.jsdelivr.net/npm/@observablehq/highlight.js@2.0.0-alpha.3/";
+const HL_ROOT =
+  "https://cdn.jsdelivr.net/npm/@observablehq/highlight.js@2.0.0/";
 
 export default function(require) {
   return function() {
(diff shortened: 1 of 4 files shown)


tmcw:
Simplified this branch a bit by undoing the usage of `async` and `import` - it no longer changes the language subset we're using in this project.

src/md.js
```diff
@@ -1,23 +1,46 @@
...
+                  .then(index => {
+                    if (index.has(block.className)) {
+                      return require(HL_ROOT +
+                        "async-languages/" +
+                        index.get(block.className)).then(language => {
+                        hl.registerLanguage(block.className, language);
+                      });
+                    }
+                  })
+                  .then(done, done);
```
visnup: Actually, what would the async/await version of this look like? Realized it's ~3 nested promises.

tmcw: It'd be a little shinier but require a change in the support expectations of this module - it's currently in a very conservative style, even using old-fashioned `var`. That's what I had initially but decided that this wasn't enough of a convenience to change the expectations of the module, and if we do, we'll want to use all the newfangled other stuff too.

visnup: ah right, forgot which module I was looking at.

PR #114, opened 2019-08-28
Remove existing fragment identifier in DOM.uid

rreusser:
Appending `"#" + id` to `window.location` results in UID hrefs with _two_ fragment identifiers. This commit removes the existing fragment identifier before appending the new one.

For an example of the currently-broken behavior, see https://observablehq.com/@mbostock/svg-clipping-test#foo. The preexisting fragment identifier `#foo` returns it to broken clipping.

### DIFF
--- a/src/dom/uid.js
+++ b/src/dom/uid.js
@@ -6,7 +6,7 @@ export default function(name) {
 
 function Id(id) {
   this.id = id;
-  this.href = window.location.href.replace(window.location.hash, "") + "#" + id;
+  this.href = new URL(`#${id}`, location) + "";
 }
 
 Id.prototype.toString = function() {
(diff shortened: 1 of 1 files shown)


src/dom/uid.js
```diff
@@ -6,7 +6,7 @@ export default function(name) {
 
 function Id(id) {
   this.id = id;
-  this.href = window.location.href + "#" + id;
+  this.href = window.location.href.replace(window.location.hash, "") + "#" + id;
```
mbostock: How about instead: 

```js
this.href = new URL(`#${id}`, location) + "";
```

rreusser: Neat. I didn't know that was a thing. That's much better!

rreusser:
```suggestion
  this.href = new URL(`#${id}`, location) + "";
```

rreusser:
Good suggestion. I've updated the PR with that version.

mbostock:
Merged into #128 for 3.2.

PR #128, opened 2019-10-18
FileAttachments: Stdlib Implementation

jashkenas:
... for use in UI, API (for downloads), and embeds.

- UI: resolve() function posts messages up to the parent frame to get signed URLs for private files.
- API: resolve() function is compiled with relative paths to downloaded files in the folder hierarchy.
- Embeds: resolve() function is compiled into the ES Modules, with absolute URLs of published files.

---

### Update

- Embeds: resolve() function is compiled into the generated ES Modules, with a map from name to absolute public (CloudFront), or private (currently CloudFront, future API) URLs.
- Downloads: resolve() function is compiled into the generated ES Modules, with a map from name to relative path on disk.
- Imports: Identical to Embeds, compiled with absolute URLs.
- UI: Passes in a `resolve()` function to this implementation that is capable of asking the parent frame (which may ask the server) for the URL for a given file.

### DIFF
--- a/src/library.js
+++ b/src/library.js
@@ -1,6 +1,5 @@
 import constant from "./constant";
 import DOM from "./dom/index";
-import FileAttachment from "./fileAttachment";
 import Files from "./files/index";
 import Generators from "./generators/index";
 import html from "./html";
@@ -18,7 +17,6 @@ export default function Library(resolver) {
   const require = requirer(resolver);
   Object.defineProperties(this, {
     DOM: {value: DOM, writable: true, enumerable: true},
-    FileAttachment: {value: constant(FileAttachment), writable: true, enumerable: true},
     Files: {value: Files, writable: true, enumerable: true},
     Generators: {value: Generators, writable: true, enumerable: true},
     html: {value: constant(html), writable: true, enumerable: true},
--- a/src/library.js
+++ b/src/library.js
@@ -1,5 +1,6 @@
 import constant from "./constant";
 import DOM from "./dom/index";
+import FileAttachment from "./fileAttachment";
 import Files from "./files/index";
 import Generators from "./generators/index";
 import html from "./html";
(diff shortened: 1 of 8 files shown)


src/library.js
```diff
@@ -17,6 +18,7 @@ export default function Library(resolver) {
   const require = requirer(resolver);
   Object.defineProperties(this, {
     DOM: {value: DOM, writable: true, enumerable: true},
+    FileAttachment: {value: constant(FileAttachment), writable: true, enumerable: true},
```
visnup: The basic implementation looks great and I think the `resolve` function is  exactly the right abstraction here. And I think stdlib is the place for this code too so that we can use it in all the places we need to use it.

That being said, I don't think we want to actually _add_ `FileAttachment` to the library here because by itself we don't want notebook code to be calling it as written. Instead, I think we should just export it from the npm module for use and in practice it'll get wrapped in the context of the situation that can provide a meaningful implementation for the `resolve` function.

jashkenas: Great idea. I was imagining that there would be cases where we would only be able to access it from the library itself ... but I guess that's not exactly true.

visnup: > we don't want notebook code to be calling it as written

Compared to all the other entries in the standard library that run as-is without _much_ further abstraction being applied on top of them. Or in other words, `FileAttachment` here isn't the exact class/function I'm getting when I call `FileAttachment` from a notebook.

yarn.lock
```diff
@@ -818,10 +818,10 @@ csstype@^2.2.0:
   resolved "https://registry.yarnpkg.com/csstype/-/csstype-2.6.6.tgz#c34f8226a94bbb10c32cc0d714afdf942291fc41"
   integrity sha512-RpFbQGUE74iyPgvr46U9t1xoQBM8T4BL8SxrN66Le2xYAPSaDJJKeztV3awugusb3g3G9iL8StmkBBXhcbbXhg==
 
-d3-require@^1.2.0:
-  version "1.2.3"
-  resolved "https://registry.yarnpkg.com/d3-require/-/d3-require-1.2.3.tgz#65a1c2137e69c5c442f1ca2a04f72a3965124c16"
-  integrity sha512-zfat3WTZRZmh7jCrTIhg4zZvivA0DZvez8lZq0JwYG9aW8LcSUFO4eiPnL5F2MolHcLE8CLfEt06sPP8N7y3AQ==
+d3-require@^1.2.4:
+  version "1.2.4"
+  resolved "https://registry.npmjs.org/d3-require/-/d3-require-1.2.4.tgz#59afc591d5089f99fecd8c45ef7539e1fee112b3"
```
mbostock: yarnpkg.com?

PR #129, opened 2019-10-28
Set crossOrigin automatically.

mbostock:


src/fileAttachment.js
```diff
@@ -31,10 +31,13 @@ class FileAttachment {
   async image() {
     const url = await this.url();
     return new Promise((resolve, reject) => {
-      const img = new Image;
-      img.onload = () => resolve(img);
-      img.onerror = () => reject(new Error(`Unable to load file: ${this.name}`));
-      img.src = url;
+      const i = new Image;
```
visnup: Out of curiosity: what was the reason to rename `img` to `i`?

mbostock: In the notebook repo it was causing prettier to wrap the line and I thought it was *prettier* to not wrap.

PR #155, opened 2020-02-19
Externalize marked.

mbostock:
Marked was internalized in 3.0.0 and significantly increased the bundle size of this library:

untitled (18)

Given that most use cases of embedding notebooks do not require Marked, I don’t think it makes sense to internalize Marked. We can still internalized Marked on our own website using d3-require rather than baking it into stdlib (and by extension, the runtime).


mbostock:
The new bundle size is 12,907 bytes, down from 29,050 (-56%).

mbostock:
FYI, I discovered I needed to expose d3-require’s default resolver to avoid duplicating the logic when inlining, so that is now exposed as Library.resolve.

PR #162, opened 2020-03-15
Error on invalid dispose function.

### LINKED ISSUE #89: Generators.observe and Generator.queue throws a TypeError on the console (side effect)
(Note: might be a duplicate of #44, but that issue doesnt specifically mention the error)

If you create a Generator with `Generators.observe` or `Generator.queue`, and that generator updates in the observable runtime, it throws on error on the console. It doesn't effect the value of the cell, the cell still computes just fine, but it seems the Observable runtime is throwing an error for some reason.

The error I get on the console (in Chrome):

```
worker.327fbbc92318b…8621ada8bfc37b.js:2 Uncaught (in promise) TypeError: i is not a function
    at Object.return (worker.327fbbc92318b…8621ada8bfc37b.js:2)
    at worker.327fbbc92318b…8621ada8bfc37b.js:2

``` 

Notebook displaying the error: https://observablehq.com/d/a28ffdc308b73051

mbostock:
Fixes #89.

My initial inclination was to just log a warning, but since warnings are only visible if you open the developer console, I figured that wouldn’t actually help people recognize the mistake.

### DIFF
--- a/src/generators/observe.js
+++ b/src/generators/observe.js
@@ -7,7 +7,9 @@ export default function(initialize) {
   const dispose = initialize(change);
 
   if (dispose != null && typeof dispose !== "function") {
-    throw new Error("invalid dispose");
+    throw new Error(typeof dispose.then === "function"
+        ? "async initializers are not supported"
+        : "initializer returned something, but not a dispose function");
   }
 
   function change(x) {
--- a/src/generators/observe.js
+++ b/src/generators/observe.js
@@ -4,7 +4,12 @@ export default function(initialize) {
   let stale = false;
   let value;
   let resolve;
-  const dispose = initialize(change);
+  let dispose = initialize(change);
+
+  if (dispose != null && typeof dispose !== "function") {
+    console.warn("dispose is not a function; ignoring", dispose);
+    dispose = null;
+  }
 
   function change(x) {
     if (resolve) resolve(x), resolve = null;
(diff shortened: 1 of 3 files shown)


visnup:
For the [example](https://observablehq.com/d/a28ffdc308b73051) linked in #89 this would instead immediately error for that cell with "invalid dispose"?

mbostock:
Saying that the returned dispose function is invalid is a hint, no?

We could have an error message that specifically targets promises by looking for a then-able if you think that’s worth the trouble.

visnup:
Re-reading the README docs on `Generators.observe` which does document the dispose return function, so it could be enough, but I don't think it's obvious that the return value is always consumed that way, especially when mixing with async.

I was thinking of the warning you get if you try to use an async function inside of useEffect:

src/generators/observe.js
```diff
@@ -6,6 +6,10 @@ export default function(initialize) {
   let resolve;
   const dispose = initialize(change);
 
+  if (dispose != null && typeof dispose !== "function") {
+    throw new Error("invalid dispose");
```
mbostock: Something like this?

```suggestion
    throw new Error(typeof dispose.then === "function"
        ? "async initializers are not supported"
        : "initializer returned something, but not a dispose function");
```

visnup: Yeah, this is greatly helpful!

PR #172, opened 2020-06-18
File attachment URL strings (3.3.2)

jashkenas:
Ref: https://github.com/observablehq/notebook/issues/5002
Previously: https://github.com/observablehq/compiler/pull/24
(Private issues)

If a FileAttachment object is created with a `URL` object, ensure that it is coerced to a string.

### DIFF
--- a/test/fileAttachments-test.js
+++ b/test/fileAttachments-test.js
@@ -6,3 +6,13 @@ test("FileAttachments is exported by stdlib", t => {
   t.equal(FileAttachments.name, "FileAttachments");
   t.end();
 });
+
+test("FileAttachment ensures that URLs are strings", async t => {
+  const fileAttachments = FileAttachments((name) =>
+    new URL(`https://example.com/${name}.js`)
+  );
+  const file = fileAttachments("filename");
+  t.equal(file.constructor.name, "FileAttachment");
+  t.equal(await file.url(), "https://example.com/filename.js");
+  t.end();
+});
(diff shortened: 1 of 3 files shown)


mbostock:
I feel it would be more appropriate to do the coercion in the FileAttachments method (rather than the constructor), since that’s where we coerce *name* to a string. Or, we could coerce both *name* and *url* to strings in the constructor, but that’s redundant with the coercion already happening in the FileAttachments function (which is also the only place that this constructor is ever called).

jashkenas:
I’ve updated the PR to move the coercion to the `url()` function instead — I looked closer at the comment that reminds us that the FileAttachments `resolve` function may resolve URLs asynchronously. We should only await that promise at the point where the `.url()` function is finally being called.

Plus a test.

test/fileAttachments-test.js
```diff
@@ -6,3 +6,23 @@ test("FileAttachments is exported by stdlib", t => {
...
+  const file = fileAttachments("filename");
+  t.equal(file.constructor.name, "FileAttachment");
+  t.equal(await file.url(), "https://example.com/filename.js");
+  t.end();
+});
+
+test("FileAttachment works with Promises that resolve to URLs", async t => {
+  const fileAttachments = FileAttachments((name) =>
+    new Promise((resolve) => resolve(new URL(`https://example.com/${name}.js`)))
+  );
```
mbostock, line 23:
```suggestion
  const fileAttachments = FileAttachments(async (name) =>
    new URL(`https://example.com/${name}.js`)
  );
```

PR #180, opened 2020-09-04
Add CSV and TSV parsing.

mbostock:
As described here: https://observablehq.com/d/8f9d08fe2416fc3f

### DIFF
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -1,7 +1,16 @@
+import {require as requireDefault} from "d3-require";
+
 async function remote_fetch(file) {
   const response = await fetch(await file.url());
   if (!response.ok) throw new Error(`Unable to load file: ${file.name}`);
   return response;
+}
+
+async function dsv(file, delimiter, {array = false, typed = false} = {}) {
+  const [text, d3] = await Promise.all([file.text(), requireDefault("d3-dsv@2.0.0/dist/d3-dsv.min.js")]);
+  return (delimiter === "\t"
+      ? (array ? d3.tsvParseRows : d3.tsvParse)
+      : (array ? d3.csvParseRows : d3.csvParse))(text, typed && d3.autoType);
 }
 
 class FileAttachment {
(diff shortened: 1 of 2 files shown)


src/fileAttachment.js
```diff
@@ -1,9 +1,17 @@
+import {require as requireDefault} from "d3-require";
+
 async function remote_fetch(file) {
   const response = await fetch(await file.url());
   if (!response.ok) throw new Error(`Unable to load file: ${file.name}`);
   return response;
 }
 
+async function dsv(file, format, {array = false, type = null} = {}) {
```
visnup, line 9: Left a comment in the notebook about `type` being a bit too generic of an argument name here. I think I'd personally prefer `typed` but that's also not a home run either.

src/fileAttachment.js
```diff
@@ -1,9 +1,17 @@
...
+  const [text, d3] = await Promise.all([file.text(), requireDefault("d3-dsv@2.0.0/dist/d3-dsv.min.js")]);
+  if (type === true) type = d3.autoType;
+  return d3[`${format}Parse${array ? "Rows" : ""}`](text, type);
```
visnup, line 12: This is a little scary since my brain went immediately to "user can specify an arbitrary function name?" but 1. both `format` and `array` aren't directly from the user, 2. it'll always be a Parse function, 3. it'll always be constrained to d3-dsv.

visnup (approved):

src/fileAttachment.js
mbostock: I’ve updated it to `typed` and added `map` as a separate argument (that can be combined with `typed`, which is great because we couldn’t do that before!).

src/fileAttachment.js
mbostock: It should be equivalent to:

```js
  return (format === "csv"
      ? (array ? d3.csvParseRows : d3.csvParse)
      : (array ? d3.tsvParseRows : d3.tsvParse))
    (text, map);
```

Would you prefer that?

visnup: Yeah, I think so. It future-proofs it a bit from someone not understanding the implications when calling this function. It's probably more legible too!

visnup:
I just realized, I'm looking forward to not having to `d3 = require('d3-dsv')` in FileAttachment-using notebooks anymore. 😄

mbostock:
As discussed, going to remove the *map* option for now to keep things simple. We can add it in the future if we want it (instead of await + map + filter).

PR #207, opened 2021-05-05
Add recommended libraries

mbostock:
This adds a handful of first-party and sanctioned third-party libraries to the Observable standard library for convenient access from notebooks. This set of libraries may grow over time based on demand. The current set is:

* `d3` - [D3.js](https://d3js.org)
* `htl` - Observable’s [Hypertext Literal](https://github.com/observablehq/htl)
* `Inputs` - [Observable Inputs](https://github.com/observablehq/inputs)
* `Plot` - [Observable Plot](https://github.com/observablehq/plot)
* `_` - [Lodash](https://lodash.com)
* `vl` - [Vega-Lite API](https://vega.github.io/vega-lite-api/)

In the last case, some additional code is required based on [this notebook](https://observablehq.com/@vega/vega-lite-api).

The trickiest part here is version pinning. I chose to hard-code the versions and paths in the standard library for stability and performance, and in particular to guarantee that duplicate versions of D3 and Hypertext Literal are not loaded: Observable Plot depends on D3, and Observable Inputs depends on Hypertext Literal. A consequence of this choice is that we’ll need to upgrade the Observable standard library explicitly, for everyone, when new versions of these libraries are released. However, when Observable supports version pinning natively in the future, we should be able to relax these requirements and give authors control over which version to use. In the meantime, notebook authors can require the libraries explicitly rather than depending on the standard library if they want to specify the version.

### DIFF
--- a/src/library.js
+++ b/src/library.js
@@ -1,5 +1,4 @@
 import {require as requireDefault} from "d3-require";
-import constant from "./constant.js";
 import DOM from "./dom/index.js";
 import Files from "./files/index.js";
 import {NoFileAttachments} from "./fileAttachment.js";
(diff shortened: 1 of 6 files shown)


shancarter:
I don't have an solutions to offer, but having both `htl` and `html` feels like a bummer. Or does one replace the other?

mbostock:
We can’t remove `html` without breaking _lots_ of notebooks. When we support standard library versioning, we can replace the existing `html` and `svg` directly instead of introducing `htl`.

src/library.js
```diff
@@ -13,24 +12,39 @@ import resolve from "./resolve.js";
...
+  Object.defineProperties(this, properties({
+    DOM: () => DOM,
+    FileAttachment: () => NoFileAttachments,
+    Files: () => Files,
+    Generators: () => Generators,
+    Inputs: () => require("@observablehq/inputs@0.7.21/dist/inputs.umd.min.js"),
+    Mutable: () => Mutable,
+    Plot: () => require("@observablehq/plot@0.1.0/dist/plot.umd.min.js"),
+    Promises: () => Promises,
+    _: () => require("lodash@4.17.21/lodash.min.js"),
```
wiltsecarpenter, line 29: This one is sort of not-like-the-others in this set. I assume that it is very commonly used and stable in notebooks?

wiltsecarpenter (approved):

src/library.js
mbostock: Lodash is the most popular library on npm according to this analysis:

https://gist.github.com/anvaka/8e8fa57c7ee1350e3491

It appears to be referenced in 2.5K public notebooks according to this query:

https://observablehq.com/search?query=lodash

That’s less than Vega (9K) and D3 (10K+), but still quite popular, and I vouch for its quality.

tophtucker:
how much does this increase the size of the bundle? we could add an example to the examples repo showing how to roll your own version-pinning for download notebooks (i.e. the example would just sub in an older or pruned version of stdlib)

mbostock:
It has a negligible increase in the bundle size because everything is loaded lazily. You only load d3 if you reference it.

tophtucker:
oh wow that's sweet

Fil:
To help the user could you add .version systematically? Currently each library has a different method (if at all): https://observablehq.com/@fil/stdlib-recommended-libraries

mootari:
> A consequence of this choice is that we’ll need to upgrade the Observable standard library explicitly, for everyone, when new versions of these libraries are released.

Is the current plan to hold out on BC breaking updates for these libraries until Observable supports version pinning out of the box?

mbostock:
@mootari Yes, we’re only going to accept patch and minor upgrades until Observable supports version pinning.

src/library.js
Martien: Question regarding evolvability and compatibility: I started out using Ramda because of its elegant “immutable auto-curried iteratee-first data-last methods”. However, including the much more popular `lodash` library conveniently and lazily by default urges me to switch to `lodash`. Plus, I just learned that there also is `lodash/fp`, that enjoys the same feature as Ramda above. Is there a way to include `lodash/fp` by default the same way as just `lodash`? Or should I fall back to an explicit require?

- https://github.com/lodash/lodash/wiki/FP-Guide
- https://observablehq.com/@mbostock/hello-lodash-fp

PR #210, opened 2021-05-13
Add @observablehq/graphviz

mbostock:
Oops, I missed this one before, but it’s super useful!

https://observablehq.com/@observablehq/graphviz


duaneatat (approved):
+1!

PR #212, opened 2021-05-22
SQLite

mbostock:
This adds [SQL.js](https://github.com/sql-js/sql.js/) to the recommended libraries, exposed as `SQLite`. In addition, it adds a _fileAttachment_.sqlite() method which returns a SQLiteDatabaseClient: an implementation of the Observable DatabaseClient interface backed by SQLite.

### DIFF
--- a/src/sqlite.js
+++ b/src/sqlite.js
@@ -46,16 +46,6 @@ export class SQLiteDatabaseClient {
 }
 
 async function exec(db, query, params) {
-  if (params !== undefined) {
-    let i = -1;
-    query = query.replace(/\?/g, () => {
-      const param = params[++i];
-      return Array.isArray(param)
-        ? new Array(param.length).fill("?")
-        : "?";
-    });
-    params = params.flat(1);
-  }
   const [result] = await db.exec(query, params);
   if (!result) return [];
   const {columns, values} = result;
--- a/src/sqlite.js
+++ b/src/sqlite.js
@@ -26,22 +26,11 @@ export class SQLiteDatabaseClient {
     return pre;
   }
   async describe(object) {
-    if (object !== undefined) {
-      const row = await this.queryRow(`SELECT * FROM '${object}' LIMIT 1`);
-      return table(
-        Object.entries(row).map(([column_name, value]) => ({
-          column_name,
-          data_type: typeof value === "string" ? "character varying"
-              : typeof value === "number" ? "integer"
-              : undefined,
-          column_default: null,
-          is_nullable: "YES"
-        }))
-      );
-    } else {
-      const rows = await this.query(`SELECT name FROM sqlite_master WHERE type = 'table'`);
-      return table(rows);
-    }
+    return table(
+      await (object === undefined
+        ? this.query(`SELECT name FROM sqlite_master WHERE type = 'table'`)
+        : this.query(`SELECT * FROM pragma_table_info(?)`, [object]))
+    );
   }
 }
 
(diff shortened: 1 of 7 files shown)


src/sqlite.js
```diff
@@ -0,0 +1,70 @@
...
+  const [result] = await db.exec(query, params);
+  if (!result) return [];
+  const {columns, values} = result;
+  const rows = values.map(row => Object.fromEntries(row.map((value, i) => [columns[i], value])));
+  rows.columns = columns;
+  return rows;
+}
+
+async function table(data, options) {
+  const Inputs = await requireDefault("@observablehq/inputs@0.8.0/dist/inputs.umd.min.js");
```
mbostock, line 68: I’m not wild about duplicating the pinned version here (ideally, it’d be managed in yarn.lock) and also it would be better to use the Library’s _require_ rather than _requireDefault_. We have the same issue with _fileAttachment_.csv, so I figured this would be acceptable as a starting point, but at some point we have to figure out how to plumb the Library’s custom _require_ through the FileAttachment function and then into the SQLiteDatabaseClient.

src/sqlite.js
```diff
@@ -0,0 +1,49 @@
+import {require as requireDefault} from "d3-require";
+
+export default async function sqlite(require) {
+  const sql = await require("sql.js@1.5.0/dist/sql-wasm.js");
+  return sql({locateFile: file => `https://cdn.jsdelivr.net/npm/sql.js@1.5.0/dist/${file}`});
+}
+
+export class SQLiteDatabaseClient {
```
visnup, line 8: Is it going to be odd that this database client doesn't extend the worker's DatabaseClient function, prototype-chain-wise, even though it implements the same interface? It only occurred to me because queryRow matches exactly the worker's implementation and could be re-used here without copying.

I was thinking alternatively, this implementation could move to the worker and it could accept a file attachment for the constructor: `DatabaseClient(FileAttachment("db.sqlite"))`. Though, there's a lot of assumption going into that expression since nothing references "sqlite" statically.

src/sqlite.js
```diff
@@ -0,0 +1,49 @@
...
+  if (!result) return [];
+  const {columns, values} = result;
+  const rows = values.map(row => Object.fromEntries(row.map((value, i) => [columns[i], value])));
+  rows.columns = columns;
+  return rows;
+}
+
+async function table(data, options) {
+  const Inputs = await requireDefault("@observablehq/inputs@0.8.0/dist/inputs.umd.min.js");
+  return Inputs.table(data, options);
```
visnup, line 48: We should switch out https://github.com/observablehq/observablehq/blob/f020251a0bc3a4cfd34e63a6684a41f0eadb7e6e/notebook-next/src/worker/databaseClient.js#L192 to use Inputs.Table too?

mbostock: I didn’t yet, but I can.

src/sqlite.js
mbostock: DatabaseClient isn’t currently part of the standard library, so I’ll need to move parts of it up here to make that work. I’ll take a crack at it.

mbostock: I thought about some more and I don’t think this is worth the effort. It’s not easy to move the DatabaseClient class into the standard library because it hard-codes the known implementations, all of which currently require coordination with the editor (since they are remote databases that require a database connector, not an in-memory local database like SQLite).

src/sqlite.js
mbostock: I decided to go the other way and port the current DatabaseClient implementation into the standard library (i.e., generate a basic HTML table rather than use Inputs.table). That’s at least consistent. I suppose we could still expose an abstract DatabaseClient here in the standard library to share some code with the other implementations, but I don’t think it’s urgent that we do this so I’d rather not, for now.

visnup (approved):

PR #215, opened 2021-05-27
XLSX file attachments

mbostock:
This adds support for XLSX file attachments via [SheetJS](https://sheetjs.com). Similar to SQLiteDatabaseClient, I also chose to wrap the SheetJS API with a more convenient, minimal abstraction.

Screen Shot 2021-05-26 at 7 45 38 PM

I chose not to expose SheetJS as a recommended library (for now). This step is not immediately needed to support XLSX file attachments, and there are multiple versions of SheetJS: we only need the “mini” version for XLSX, but we’d need the much bigger “full” version for XLS.

### DIFF
--- a/src/xlsx.js
+++ b/src/xlsx.js
@@ -1,0 +1,20 @@
+let XLSX; // set lazily
+
+export default async function xslx(require) {
+  return XLSX = await require("xlsx@0.17.0/dist/xlsx.mini.min.js");
+}
+
+export class XlsxWorkbook {
+  constructor(workbook) {
+    Object.defineProperties(this, {
+      _: {value: workbook}
+    });
+  }
+  sheetNames() {
+    return this._.SheetNames.slice();
+  }
+  sheet(name, options) {
+    if (!Object.prototype.hasOwnProperty.call(this._.Sheets, name)) throw new Error("unknown sheet");
+    return XLSX.utils.sheet_to_json(this._.Sheets[name], options);
+  }
+}
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -2,7 +2,7 @@
   "extends": "eslint:recommended",
   "parserOptions": {
     "sourceType": "module",
-    "ecmaVersion": 8
+    "ecmaVersion": 9
   },
   "env": {
     "browser": true,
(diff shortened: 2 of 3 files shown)


src/xlsx.js
```diff
@@ -0,0 +1,20 @@
+let XLSX; // set lazily
+
+export default async function xslx(require) {
+  return XLSX = await require("xlsx@0.17.0/dist/xlsx.mini.min.js");
+}
+
+export class XlsxWorkbook {
+  constructor(workbook) {
+    Object.defineProperties(this, {
+      _: {value: workbook}
```
visnup, line 10: It'd be nice to expose the XLSX or XLSX.utils somewhere so people can easily use those functions like sheet_to_html or sheet_to_formulae. Or for modifying and writing back out to xlsx.

mbostock: Right, I commented on why I didn’t do this: there are two versions of XLSX (mini and full), and we only want the mini one to parse XLSX files, but I don’t think that one should be the one in standard library. I suppose I could hide it in the XlsxWorkbook class I made, but I figured it would be better to expose a minimal API that is easier for us to improve over time. (In my opinion, the design of SheetJs is not a little weird… I’m not sure why it favors global static functions over methods.)

visnup: ok, let's start conservatively and see if we need to add more. there's also always the workaround of notebook-space requiring xlsx/sheetjs explicitly and using it if you need to.

visnup (approved):

visnup:
I think `mobile.sheet("Data", {range: 2 /* or 3? */})` may be what I was looking for.

visnup:
This is blocked on deciding what to do with dates and times and specifically time zones. We want to translate dates and times to JavaScript Date objects. SheetJS and XLSX both specify "dates and times are in your local time zone." While that might make sense from a "single file on someone's computer" perspective, it leads to a lot of inconsistency in a web environment.

CobusT:
> SheetJS and XLSX both specify "dates and times are in your local time zone."

What happens with csv attachments for dates and times? Does that have the same issue?

mbostock:
> What happens with csv attachments for dates and times? Does that have the same issue?

By default all fields are strings and no dates are parsed. But if you use d3.autoType dates are assumed to be UTC.

mbostock:
Superseded by #248.

PR #216, opened 2021-05-28
Automate version resolution

mbostock:
I wanted to make it easier to keep these versions up-to-date, and to avoid hard-coding the versions in multiple places. I couldn’t figure out a good way to track the versions using yarn (I considered devDependencies), but this at least gives us a script we can run to refresh the dependency versions and then see if we want to update.

### DIFF
--- a/bin/resolve-dependencies
+++ b/bin/resolve-dependencies
@@ -1,0 +1,73 @@
+#!/usr/bin/env node
+
+const fetch = require("node-fetch");
+
+(async () => {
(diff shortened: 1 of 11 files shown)


mbostock:
Gentle ping on the review, @visnup?

visnup (approved):
Do you need yarn to actually do work given the version or are you just looking for a convenient place to list package names and versions? I think npm and yarn are neutral when it comes to additional package.json keys and values that aren't standard so it could be possible to list the dependencies in there under a custom key... that's probably not buying us much though?

PR #222, opened 2021-06-02
Add support for ZIP file attachments.

mbostock:
Adds *fileAttachment*.zip which returns a promise to a ZipArchive. The ZipArchive exposes an *archive*.fileNames() method which returns the array of file names, and an *archive*.file(*name*) method which returns a FileAttachment-like object. So, to load a CSV file in a ZIP file, you can now say:

```js
archive = FileAttachment("ACS_14_5YR_B01003.zip").zip()
```
```js
data = archive.file("ACS_14_5YR_B01003_with_ann.csv").csv({typed: true})
```

As with XLSX #215, I’ve decided not to expose JSZip as a recommended library, and instead wrap it with a more minimal abstraction that feels more Observable-ish.

Screen Shot 2021-06-01 at 6 47 28 PM

### DIFF
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -98,11 +98,11 @@ export class ZipArchive {
   file(path) {
     const object = this._.file(path);
     if (!object || object.dir) throw new Error("file not found");
-    return new ZipFile(object);
+    return new ZipArchiveEntry(object);
   }
 }
 
-class ZipFile extends AbstractFile {
+class ZipArchiveEntry extends AbstractFile {
   constructor(object) {
     super(object.name);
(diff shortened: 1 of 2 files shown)


visnup (approved):
I assume for archives with directories, the filenames array will contain values like "dir/file.csv"? and only contain files and not "dir/"?

now thinking we could use a .tarball version.

mbostock:
> I assume for archives with directories, the filenames array will contain values like "dir/file.csv"? and only contain files and not "dir/"?

Good catch. It did include directories, but now I’ve filtered them out.

visnup (approved):

PR #223, opened 2021-06-02
Add Apache Arrow as a recommended library and supported file attachment

mbostock:
Apache Arrow is exposed as `Arrow` in the standard library, and _fileAttachment_.arrow() returns a Promise to an Arrow.Table.

Screen Shot 2021-06-02 at 3 56 19 PM

### DIFF
--- a/bin/resolve-dependencies
+++ b/bin/resolve-dependencies
@@ -60,6 +60,10 @@ const fetch = require("node-fetch");
     const package = await resolve("vega-lite-api");
     console.log(`export const vegaliteApi = dependency("${package.name}", "${package.version}", "${package.export}");`);
   }
+  {
+    const package = await resolve("apache-arrow@4");
+    console.log(`export const arrow = dependency("${package.name}", "${package.version}", "${package.export}");`);
+  }
 })();
 
 async function resolve(specifier) {
--- a/src/dependencies.js
+++ b/src/dependencies.js
@@ -13,3 +13,4 @@ export const sql = dependency("sql.js", "1.5.0", "dist/sql-wasm.js");
 export const vega = dependency("vega", "5.20.2", "build/vega.min.js");
 export const vegalite = dependency("vega-lite", "5.1.0", "build/vega-lite.min.js");
 export const vegaliteApi = dependency("vega-lite-api", "5.0.0", "build/vega-lite-api.min.js");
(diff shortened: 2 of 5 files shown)


visnup:
We're effectively pinned to these library versions, at least major if not minor versions, until we can version the standard library, correct? It feels like they are making steady, significant progress on this library and it's gone from [v1 to v4 in the past year](https://github.com/apache/arrow/releases). All of the other libraries we've recently added I assumed were pretty stable, but I'm half worried Arrow could go from 4 to 7 before we can follow suit.

mbostock:
We’re committed to backwards compatibility until we ship version pinning, yes. I don’t think we should block adding useful functionality on us shipping version pinning: I’d rather include a slightly out-of-date version of Apache Arrow in the box than nothing.

mbostock:
Side note, but it does look like 5.0.0 is already planned per the [package.json](https://github.com/apache/arrow/blob/bbf0e674f8b237f85c479680738b2fa20da6a5f0/js/package.json#L108), but I can’t find any release notes, so I’m not really sure what’s different. In any case, I think we should still go ahead, but also redouble our efforts to ship version pinning.

domoritz:
All Arrow packages are released every three months and every time there is a new major version. Note that the binary format is not changing. See https://arrow.apache.org/docs/format/Versioning.html for details.

In the past few versions, the JS library hasn't changed much but for v5, we started some significant improvements to make the library leaner and more tree-shakeable. One breaking change we already added is https://github.com/apache/arrow/pull/10277 and another significant change will be https://github.com/apache/arrow/pull/10371.

visnup (approved):
Can be updated to 4.0.1. But yeah ok, let's recommend Arrow!

mbostock:
Arrow 5.0.0 is out already, but since we plan on adding Arquero imminently, I figure we should stick to 4.0.1.

domoritz:
The api hasn't changed much between 4 and 5. The biggest change is that tables don't extend data frame anymore (but data frames still extend tables).

visnup:
Is there a reason we can't do 5 then? Would it be incompatible with the current version Arquero?

domoritz:
I don't think so but it would be good to confirm by updating arquero to v5.

mbostock:
I will investigate upgrading to Arrow 5 at the time we add Arquero.

PR #238, opened 2021-08-01
resolve jszip

mbostock:
I missed this when rebasing #222. This also upgrades JSZip from 3.6.0 to 3.7.0 which you can see here:

https://github.com/Stuk/jszip/compare/v3.6.0...e5b3f0ddaa8182cd6ea253e97f678b9f36d0d8ac

The only change appears to be a security fix for weirdly-named files which I do not think applies in our usage.

### DIFF
--- a/bin/resolve-dependencies
+++ b/bin/resolve-dependencies
@@ -1,6 +1,8 @@
 #!/usr/bin/env node
 
 const fetch = require("node-fetch");
+const extensionRe = /\.[^/]*$/;
+const mains = ["unpkg", "jsdelivr", "browser", "main"];
 
 (async () => {
   console.log(`import dependency from "./dependency.js";`);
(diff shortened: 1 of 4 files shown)


bin/resolve-dependencies
```diff
@@ -72,6 +78,17 @@ async function resolve(specifier) {
   return {
     name: package.name,
     version: package.version,
-    export: (package.unpkg || package.jsdelivr || package.browser || package.main).replace(/^\.\//, "")
+    export: main(package)
```
mbostock, line 81: This is a workaround for JSZip’s use of the nonstandard `browser` object for Browserify:

https://github.com/Stuk/jszip/blob/e5b3f0ddaa8182cd6ea253e97f678b9f36d0d8ac/package.json#L27-L30

I modeled this code after the same code from d3-require.

visnup (approved):

PR #243, opened 2021-08-22
inline d3-dsv

mbostock:
This inlines the d3-dsv dependency which is needed to support CSV file attachments. This adds 1,019 bytes to the gzipped file size (from 6,263 to 7,282), which should be negligible, and avoids an additional server roundtrip before loading a CSV file.


mbostock:
Note that I stuck with d3-dsv@2 since d3-dsv@3 adopts ES modules, which would require this package to adopt ES modules as well as a major version change. I think we should do that in the near future, but it doesn’t need to be part of this change (and there are no other meaningful changes in d3-dsv@3 yet that we’re missing out on).

mbostock:
Superseded by #244.

PR #248, opened 2021-09-03
XLSX support with ExcelJS

visnup:
Alternative to #215 

Handles dates correctly, using UTC. Based largely on @Fil's previous exploration.
- [x] Test suite
- [x] Coerce formula errors to NaN
- [x] Escape the hyperlink case properly
- [x] Require ":" in string ranges (error if missing) to choose the meaning of "A" as either "A:" or "A:A", explicitly
- [x] Row numbers

Screen Shot 2021-09-10 at 3 45 19 PM

__The main goal of this PR is to give people a way to extract _data_ out of xlsx files quickly, efficiently, and correctly.__ To do that, we assume people should be able to _visually recognize and find_ the data they want to extract, leveraging any previous familiarity with the xlsx file. We don't want to spend much effort on preserving styling or presentation (widths, fonts, value formatting, merged cells, frozen panes) or features used during building a spreadsheet like formula definitions (only results are extracted). At the same time, we might decide some presentation features are worth preserving _if we think it would help people trust and recognize_ the extracted contents of the spreadsheet (number formatting I'd guess could fall into this).

Also, we should make it easy to maintain extracting new or updated data from an updated or mutable file, which implies the importance of the unbounded range feature ("12:" to mean extract starting at row 12 to the end).

The extracted data should work well with the rest of the downstream toolchains in Observable. So, plain JavaScript values (`NaN`) over descriptive objects (`{ error: "#DIV/0" }`).

The `Workbook` API hopefully will be reusable to represent a Google Sheet just as well in the future.

The assumed, frequent workflow of this API is that a user would pass all of the data first to `Inputs.table` or similar for exploratory recognition, then filter it down using the `range` and `headers` options described below.

## _Workbook.sheet(name, { range, headers }): Record<string, any>[]_

Returns an array of objects representing the contents of cells in a specific sheet of the workbook. An example return value may look something like:

```javascript
[ { A: "Data were collected and made available by Dr. Kristen Gorman and the Palmer Station, Antarctica LTER, a member of the Long Term Ecological Research Network." },
  { A: "Region", B: "Island", C: "Date Egg", D: "Culmen Length (mm)" },
  { A: "Anvers", B: "Torgersen", C: 2007-11-11, D: 39.1 },
  ...
]
```

Empty cells are skipped: objects will not include fields or values for them. Empty rows are _not_ skipped assuming they'll aid in data recognition. Values are coerced to their JavaScript types: numbers, strings, Date objects. Formula results are included, but formula definitions ignored. Row numbers from the source sheet are included to assist with range specification and recognition.

___name: string | number___ is a sheet name to get data for. If it's a string, it must match a name in `Workbook.sheetNames`. You can pass a zero-indexed number to get the corresponding sheet in order of `Workbook.sheetNames`. For example, `sheet(0)` is the first sheet.

___range: string___ specifies a single rectangular range of cells to extract from the sheet as an Excel-based representation of a range, e.g. `"B4:L123"` to mean from cell B4 in the top left to L123 in the bottom right, inclusive. By default if no range is specified, the entire sheet is extracted.

Similar to Excel, the row or column part of the start or end may be omitted to mean the entire row or column, e.g. `"4:123"` to mean rows 4 through 123 inclusive. Extending the standard syntax, you may omit a start or end specifier to mean "A1" or last column and last row, e.g. `"4:"` to mean row 4 to the end of the sheet.

Union `"A1:B3,D1:G3"` and intersection `"A1:C3 B2:D4"` specifiers are not supported.

___headers: boolean___ will treat the first extracted row as column headers and use their values as field names for returned objects. The default is false. If a value doesn't exist in the header row for a value, column names (A-ZZ) will be used instead. Underscores (_) are appended if field names are repeated.

With `{ range: "2:", headers: true }`, the above penguins data would be:

```javascript
[ { Region: "Anvers", Island: "Torgersen", "Date Egg": 2007-11-11, "Culmen Length (mm)": 39.1 },
  ...
]
```


src/xlsx.js
```diff
@@ -0,0 +1,100 @@
...
+  let s = "";
+  x++;
+  do {
+    s = String.fromCharCode(64 + (x % 26 || 26)) + s;
+  } while ((x = Math.floor((x - 1) / 26)));
+  return s;
+}
+
+function NN(s = "") {
+  const [, sc, sr] = s.match(/^([a-zA-Z]+)?(\d+)?$/);
```
visnup, line 92: Are these optional groups dangerous from a regex backoff standpoint?

src/fileAttachment.js
```diff
@@ -70,6 +71,10 @@ class AbstractFile {
   async html() {
     return this.xml("text/html");
   }
+  async xlsx() {
+    const [ExcelJS, buffer] = await Promise.all([requireDefault(exceljs.resolve()), this.arrayBuffer()]);
+    return new ExcelWorkbook(await new ExcelJS.Workbook().xlsx.load(buffer));
```
mbostock, line 76: I’m not sure if there’s really an advantage to this, but it _is_ possible to stream…

```suggestion
    const [ExcelJS, stream] = await Promise.all([requireDefault(exceljs.resolve()), this.stream()]);
    return new ExcelWorkbook(await new ExcelJS.Workbook().xlsx.read(readStream(stream)));
```

where

```js
async function* readStream(stream) {
  const reader = stream.getReader();
  try {
    while (true) {
      const {done, value} = await reader.read();
      if (done) return;
      yield value;
    }
  } finally {
    reader.releaseLock();
  }
}
```

That said, it feels like purely a novelty, so unless you know of a reason to favor streaming I would stick with this approach and just leave this comment for history in case we’d like to revisit.

visnup: This might be better on memory usage for large files, but unknown how much better. It'd be interesting to find some realistically large files and see what performance is really like.

visnup:
Should `fileAttachment.xslx()` optionally take the same arguments as `ExcelWorkbook.sheet(name, options)` as a kind of shorthand? And if you pass them, it calls `.sheet` for you and returns that value? It would make the case of extracting a specific sheet and range a one-liner…

test/xlsx-test.js
```diff
@@ -0,0 +1,156 @@
...
+      Sheet1: [
+        ["one", "one", "one", "two"],
+        [1, null, 3, 4],
+        [5, 6, 7, 8],
+      ],
+    })
+  );
+  t.same(workbook.sheet(0, {headers: true}), [
+    {one: 1, one_: 3, two: 4},
+    {one: 5, one__: 6, one_: 7, two: 8},
```
Fil, line 75: This one is particularly sneaky; the layout of keys gets inverted.

src/xlsx.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+function valueOf(cell) {
+  if (!cell) return;
+  const {value} = cell;
+  if (value && typeof value === "object") {
+    if (value.formula) return value.result;
+    if (value.richText) return value.richText.map((d) => d.text).join("");
+    if (value.text && value.hyperlink)
+      return `<a href="${value.hyperlink}">${value.text}</a>`;
```
Fil, line 53: might hyperlink contain chars that should be escaped (like a double quote)? or text contain a `</a>`?

test/xlsx-test.js
visnup: agreed. I think maybe if there's a header row I can seed all of the column names using that row?

src/xlsx.js
visnup: not really sure how to represent HTML here. should I reuse the html tagged template literal?

Fil (commented):
Code review and suggested changes: https://github.com/observablehq/stdlib/pull/249

Here's a few more questions I'd have:
* develop ExcelWorkbook in README.md?
* discussion on https://observablehq.com/d/25d78559efffa7fb#comment-c830048f97493242
* can we have {text: xxx} and no hyperlink property?
* escaping html (in links)
* add a few large files to test?
* name: xlsx; ExcelWorkbook, or Spreadsheet?
* confusion on the rows’ numbers (they start at 1 in the traditional string specifier, at 0 in array specifier)

src/xlsx.js
```diff
@@ -0,0 +1,99 @@
+export class ExcelWorkbook {
+  constructor(workbook) {
+    Object.defineProperty(this, "_", {value: workbook});
+  }
+  sheetNames() {
+    return this._.worksheets.map((sheet) => sheet.name);
+  }
```
mbostock, line 7: Can we make this an enumerable property rather than a method, so that it’s visible in the inspector and feels more consistent with ZipArchive?

https://github.com/observablehq/stdlib/blob/466685ddb0eb60c2192b550b5bc57d5ed0018baa/src/fileAttachment.js#L100-L110

src/xlsx.js
```diff
@@ -0,0 +1,99 @@
...
+    const _row = sheet._rows[r];
+    if (!_row || !_row.hasValues) continue;
+    const row = (output[r - r0] = {});
+    for (let c = c0; c <= c1; c++) {
+      const value = valueOf(_row._cells[c]);
+      if (value !== null && value !== undefined) row[name(c)] = value;
+    }
+  }
+
+  output.columns = names.filter(() => true);
```
mbostock, line 43: What is this doing?

src/xlsx.js
```diff
@@ -0,0 +1,99 @@
...
+    if (!names[n]) {
+      let name = (headerRow ? valueOf(headerRow._cells[n]) : AA(n)) || AA(n);
+      while (seen.has(name)) name += "_";
+      seen.add((names[n] = name));
+    }
+    return names[n];
+  }
+  if (headerRow) for (let c = c0; c <= c1; c++) name(c);
+
+  const output = new Array(r1 - r0 + 1).fill({});
```
mbostock, line 32: This is filling the output with a shared empty object for all rows, whereas the rows with values are reassigned below to new objects. Do we want to use an empty object to represent rows without values, rather than undefined? If we do want to use an empty object, I think we’ll still want a distinct object for each row, rather than sharing the object across rows. That could be done by moving the `output[r - r0] = {}` below before the `continue` rather that using array.fill here.

src/xlsx.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+}
+
+function extract(sheet, {range, headers}) {
+  let [[c0, r0], [c1, r1]] = parseRange(range, sheet);
+  const seen = new Set();
+  const names = [];
+  const headerRow = headers && sheet._rows[r0++];
+  function name(n) {
+    if (!names[n]) {
```
mbostock, line 23:
```suggestion
    if (!(n in names)) {
```

This avoids the chance that the name could be a falsely value (e.g., the empty string).

src/xlsx.js
Fil: The goal is to add a `columns: ["A", "B", …]` property to the data array, like d3.csv and FileAttachment().csv().

The filtering… maybe a placeholder for removing all the columns that have no value in the dataset? 🤔

src/xlsx.js
Fil: I don't think we'd want the empty string as a column name, it's a case where I'd rather replace it with the column’s AA value.

src/xlsx.js
visnup: The filtering is compacting the sparseness. Could be written more clearly as `names.filter((d) => d !== undefined)`, but `filter` by itself skips over the sparse gaps. And yeah, adds the `columns` property so that `Inputs.table` will display the columns in the correct order as row object fields could come in sparsely/out of order.

src/xlsx.js
visnup: I think I actually tried to simply use sparseness initially but `Inputs.table` had trouble with it? re-using the same object was a pre-optimization from a memory usage standpoint. I also toyed with using a `Symbol(empty)` to make it even more explicit.

I slightly want to auto-filter these rows out of the return value since I feel like in usage that would be one of the first things I'd always end up writing anyway in the notebook, but that seemed like possibly surprising behavior at the same time?

src/xlsx.js
visnup: hm, what about `"0"`?

src/xlsx.js
visnup: Why writable?

src/xlsx.js
visnup: properly escaped the string. debated bringing in [escape-html](https://www.npmjs.com/package/escape-html), but wrote my own function. 🤔

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+        enumerable: true,
+      },
+    });
+  }
+  sheet(name, {range, headers = false} = {}) {
+    const sname =
+      typeof name === "number"
+        ? this.sheetNames[name]
+        : this.sheetNames.includes(name + "")
+        ? name + ""
```
mbostock, line 17: Nit: Coerce once to a string once to avoid potentially nondeterministic toString?

```suggestion
        : this.sheetNames.includes(name += "")
        ? name
```

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+    Object.defineProperties(this, {
+      _: {value: workbook},
+      sheetNames: {
+        value: workbook.worksheets.map((sheet) => sheet.name),
+        writable: false,
+        enumerable: true,
+      },
+    });
+  }
+  sheet(name, {range, headers = false} = {}) {
```
mbostock, line 12: Nit: Since these options aren’t used here, pass them through?

```suggestion
  sheet(name, options) {
```

README.md
```diff
@@ -379,6 +379,15 @@ Returns a promise to the file loaded as a [SQLite database client](https://obser
 const db = await FileAttachment("chinook.db").sqlite();
 ```
 
+<a href="#attachment_xlsx" name="attachment_xlsx">#</a> *attachment*.<b>xlsx</b>() [<>](https://github.com/observablehq/stdlib/blob/master/src/xlsx.js "Source")
+
+Returns a promise to the file loaded as a [Workbook](https://observablehq.com/@observablehq/xlsx-workbook).
```
mbostock, line 384: I shortened the URL of the notebook for memorability.

```suggestion
Returns a promise to the file loaded as a [Workbook](https://observablehq.com/@observablehq/xlsx).
```

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+    return [
+      [c0, r0],
+      [c1, r1],
+    ];
+  } else {
+    throw new Error("Unknown range specifier");
+  }
+}
+
+function AA(c) {
```
mbostock, line 112: What does this function do? The name AA is cryptic. Some comments would be nice.

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+  let sc = "";
+  c++;
+  do {
+    sc = String.fromCharCode(64 + (c % 26 || 26)) + sc;
+  } while ((c = Math.floor((c - 1) / 26)));
+  return sc;
+}
+
+function NN(s) {
```
mbostock, line 121: What does this function do? The name NN is cryptic. Some comments would be nice.

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+  const [, sc, sr] = s.match(/^([A-Z]*)(\d*)$/i);
+  let c = undefined;
+  if (sc) {
+    c = 0;
```
mbostock, line 125: Nit: Simpler?

```suggestion
  let c = 0;
  if (sc) {
```

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+    ];
+  } else if (typeof specifier === "object") {
+    const [
+      [c0 = 0, r0 = 0] = [],
+      [c1 = columnCount - 1, r1 = rowCount - 1] = [],
+    ] = specifier;
+    return [
+      [c0, r0],
+      [c1, r1],
+    ];
```
mbostock, line 106: Do we need to support an object form as well as a string form for _specifier_? Is the string form (e.g., `B4:AF234`) the more familiar of the two? I’d guess so, so I’d favor just having the string form here unless there’s a strong precedent.

In addition to being simpler, here are two more reasons to use only a string. First, it’s a simpler structure that is easier to validate. For example, I can crash the code with `TypeError: null is not an object` by saying `parseRange([null, null], {})` or `parseRange(null, {})` and it’d be better to have a nice error message. Second, it lets us coerce the specifier to a string so that objects that implement toString can be used as range specifiers if desired. This is a nitch feature to be sure, but idiomatic JavaScript given how nearly everything is typeof object. (Plot has a more specific test for object literals [using toString](https://github.com/observablehq/plot/blob/8b47987bbc52ec08d897c3bf39af26a1b5835d85/src/mark.js#L186-L192). You could also use Array.isArray here on the specifier as a more specific check, but my point is it’s better to avoid type checking at all if possible.)

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+  let sc = "";
+  c++;
+  do {
+    sc = String.fromCharCode(64 + (c % 26 || 26)) + sc;
+  } while ((c = Math.floor((c - 1) / 26)));
+  return sc;
+}
+
+function NN(s) {
+  const [, sc, sr] = s.match(/^([A-Z]*)(\d*)$/i);
```
mbostock, line 122: This will crash if `s` doesn’t match the expected form (e.g., `N("-")`) because s.match will return null. You can reproduce this by saying `parseRange("-:-", {})`. It might be okay to trust that ExcelJS always returns values that are compatible with NN, but if we’re passing a user-specified range specifier to NN, we should protect (possibly in parseRange rather than here) against invalid input to produce a better error message.

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+        return "&quot;";
+      case "'":
+        return "&#39;";
+    }
+  });
+}
+
+function parseRange(specifier = [], {columnCount, rowCount}) {
+  if (typeof specifier === "string") {
+    if (!specifier.includes(":")) throw new Error("Malformed range specifier");
```
mbostock, line 91: I think we want to more strict here about how we expect the specifier to be formatted. If we’re lax, then we’re likely to get a more confusing error downstream, e.g., `parseRange("-:-", {})`. Maybe a regex?

src/xlsx.js
```diff
@@ -0,0 +1,130 @@
...
+  if (value && value instanceof Date) return value;
+  if (value && typeof value === "object") {
+    if (value.formula || value.sharedFormula)
+      return value.result && value.result.error ? NaN : value.result;
+    if (value.richText) return value.richText.map((d) => d.text).join("");
+    if (value.text)
+      return value.hyperlink
+        ? `<a href="${encodeURI(value.hyperlink)}">${escapeHTML(
+            value.text
+          )}</a>`
```
mbostock, line 65: We’re losing type information here: the caller won’t be able to distinguish between plain text and HTML. This gets to my “big question” at the top of the review. Some possible alternatives: return an HTML element;  ignore the hyperlink and return the text; return the object returned by ExcelJS rather than a simplified representation.

mbostock (commented):
Exciting to see this come together! One big thing, and a lot of little nits. (Don’t worry about the nits too much.)

The big thing: documentation and context/motivations for this API are needed for this review. For example: How is the *range* option specified? Why do we support both a string form and an object form of ranges? How is the *header* option implemented (_i.e._, how do we define header rows)? How are formulas, rich text, and hyperlinked values represented, and why? _etc._

This is a nontrivial abstraction on top of ExcelJS, so I want to understand our goals. I can try to guess intent from reading the source code, but I imagine you’ve been thinking about this a bunch already and I would like to hear your perspective. And if you write it down, it’ll give us a headstart on the documentation we eventually share with users.

Thanks!

visnup:
> The big thing: documentation and context/motivations for this API are needed for this review. For example: How is the _range_ option specified? Why do we support both a string form and an object form of ranges? How is the _header_ option implemented (_i.e._, how do we define header rows)? How are formulas, rich text, and hyperlinked values represented, and why? _etc._

Good point. I took a first stab at motivation and goals and documentation of `sheet()` in the description. Will follow up on the other comments later.

src/xlsx.js
visnup: Yeah, I think you're right and we should drop the array form. Pretty sure the more useful, efficient, and friendly form is the string one. The array one I think I only included for the spreadsheet-averse people.

Fil: The array form can be useful for someone who would create a UI (say, with sliders or a brush) to specify the range. (But we could say they'd just have to copy the AA function).

src/xlsx.js
Fil: AA takes a column number *c* and returns a column name *sc*; e.g. 0 becomes "A", 26 is "AA".

src/xlsx.js
Fil: Worse: NN is not the inverse of AA, because it's working on "A1" to return [0, 0], "A" to return [0, undefined] and "1" to return [undefined, 0]. We could clarify the functions names to be:
- string columnName(int: )
- int columnNumber(string: )

and probably do the same for the rows ("1" <=> 0)?

README.md
```diff
@@ -379,6 +379,15 @@ Returns a promise to the file loaded as a [SQLite database client](https://obser
...
+
+```js
+const workbook = await FileAttachment("profit-and-loss.xlsx").xlsx();
+const sheet = workbook.sheet("Sheet1", { range: "B4:AF234", headers: true });
```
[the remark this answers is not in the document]
visnup: hrm I need to see what's up with my vscode prettier settings since this is specified in .prettierrc.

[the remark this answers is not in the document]
visnup: ah, readme vs. code.

src/xlsx.js
visnup: Hm, I felt since this is the public API boundary, we should be defensive here about the contents of options. At the same time, I lament the `f(...)` formatting this causes the inspector to fallback to and obscure the more useful signature in a notebook.

src/xlsx.js
visnup: I know of two common cases for hyperlinks in Excel:

1. As documentation to link to more information
2. As data in a row

Simplifying this to return a simple string and also not to duplicate information when the hyperlink and text are identical. For the "data in a row" case, this feels the least obtrusive and most natural.

But your original point still stands: we're still losing that the text explicitly contains a hyperlink and the user will need to re-infer it's a URL.

src/xlsx.js
visnup: I renamed both of these and added some comments for clarity.

visnup:
@mbostock @Fil ok, this is ready for re-review. should be pretty close or final?

visnup:
To add an idea while talking to @mbostock: in the future we could offer another option `raw: true` that would return cell values which were Objects, but with `valueOf` and `toString` methods which would coerce them back to the more primitive values we're returning here. That would give people an intermediate method of getting at more of the stored information without having to use ExcelJS or SheetJS directly.

Fil (commented):
Really great. Scoped to just what's needed. 

I particularly like the hidden column "#" (row reference) which helps to refine the range interactively.

One last question: I wonder if we should clean up the cells we use as headers? I've tested this on xlsx files found at random on the web, and for example in
https://www.un.org/securitycouncil/sites/www.un.org.securitycouncil/files/cross-cutting_poc.xlsx one header is `.........Formal agenda item` (with many spaces, represented here with dots) and another one also has a linefeed:
~~~
 PP/OP/
PRST
~~~

it's usable (just a bit awkward), but if we want to do this clean-up, better earlier than later.

Fil:
my preference would be to replace all \n \r by spaces, and trim

visnup:
> my preference would be to replace all \n \r by spaces, and trim

What about combine multiple whitespace characters inside the string into a single space?

visnup:
Does d3-dsv do any type of trimming?

Fil:
no… ok, then 8-)

Fil (approved):

visnup:
> no… ok, then 8-)

yeah ok, I think the default response is to encourage people to fix sources upstream and clean up those things in the xlsx files, then everyone benefits. I noticed a few values which could use trimming too and was tempted to do it in stdlib, so it's a slippery slope.

visnup:
Can add later, but an idea re: the "#" field: I've been tempted to add a `widths` property on the returned array for `Inputs.table` to use. We could do it based on the actual widths in the spreadsheet too.

mootari:
> my preference would be to replace all \n \r by spaces, and trim

Why do you want to remove line breaks? Personally I've used them in headers quite a few times to keep long headers more readable/organized. (Sorry if this was already discussed.)

PR #254, opened 2021-09-15
Mbostock/xlsx tweaks

mbostock:


src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
   }
 }
 
-function extract(sheet, {range, headers = false} = {}) {
+function extract(sheet, {range, headers} = {}) {
```
mbostock, line 24: Since we only care whether _headers_ is truthy, there’s no advantage to a default value.

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
   let [[c0, r0], [c1, r1]] = parseRange(range, sheet);
-  const headerRow = headers && sheet._rows[r0++];
+  const headerRow = headers ? sheet._rows[r0++] : null;
```
mbostock, line 26: Since _headerRow_ is a variable we create (not an argument), we should prefer a strict value (null) rather than propagating an argument of unknown value.

src/xlsx.js
```diff
@@ -52,14 +53,16 @@ function extract(sheet, {range, headers = false} = {}) {
 function valueOf(cell) {
   if (!cell) return;
   const {value} = cell;
-  if (value && value instanceof Date) return value;
-  if (value && typeof value === "object") {
-    if (value.formula || value.sharedFormula)
+  if (value && typeof value === "object" && !(value instanceof Date)) {
```
mbostock, line 56: Here I’m just reducing the number of equivalent branches.

src/xlsx.js
```diff
@@ -52,14 +53,16 @@ function extract(sheet, {range, headers = false} = {}) {
...
       return value.result && value.result.error ? NaN : value.result;
-    if (value.richText) return value.richText.map((d) => d.text).join("");
+    }
+    if (value.richText) {
+      return richText(value);
+    }
     if (value.text) {
       let {text} = value;
-      if (text.richText) text = text.richText.map((d) => d.text).join("");
+      if (text.richText) text = richText(text);
```
mbostock, line 65: When would you have a value = {text: {richText: […]}} instead of {richText: […]}?

visnup: I found this example in a spreadsheet from Ramona. I didn't really question all of the different richText permutations at that point, but this implies we should try to find more examples to make sure we cover all of them. I'd consider any new ones we find bugs to fix though.

mbostock: Maybe hyperlinked richText is {hyperlink: …, text: {richText: […]}} (instead of {hyperlink: …, richText: […]})?

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
 
-function extract(sheet, {range, headers = false} = {}) {
+function extract(sheet, {range, headers} = {}) {
   let [[c0, r0], [c1, r1]] = parseRange(range, sheet);
-  const headerRow = headers && sheet._rows[r0++];
+  const headerRow = headers ? sheet._rows[r0++] : null;
   let names = new Set(["#"]);
   for (let n = c0; n <= c1; n++) {
-    let name = (headerRow ? valueOf(headerRow._cells[n]) : null) || toColumn(n);
+    let name = (headerRow ? valueOf(headerRow._cells[n]) : null) || toColumn(n); // should a falsey header value be ignored? or just nullish?
```
visnup, line 29: I think falsey, and specifically I'm thinking of ""¹ because I'm pretty sure with editing and re-editing someone could see a cell as empty in Excel, but the underlying value would be "". And with that, I could see wanting to maybe also ignore all-whitespace values here too.


¹ correct me if I'm wrong that that's not falsey.

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
     while (names.has(name)) name += "_";
     names.add(name);
   }
   names = new Array(c0).concat(Array.from(names));
 
   const output = new Array(r1 - r0 + 1);
   for (let r = r0; r <= r1; r++) {
-    const row = (output[r - r0] = Object.defineProperty({}, "#", {
+    // Should we be using Object.create(null) instead of an empty object here?
+    const row = (output[r - r0] = Object.defineProperty({}, "#", { // what is this non-enumerable row["#"] property for?
```
visnup, line 38: For aid in range specification when using with Inputs.table in a notebook. The "#" column is included in the list of column names, so it shows up for every row. And even if Inputs.table gets a row number display in the future, after using the `range` option to skip an initial set of rows, it could never know what the source row number would have been.

132925195-8ede0efb-e6b5-4cc9-ac8b-ca5ba9a27774

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
+    let name = (headerRow ? valueOf(headerRow._cells[n]) : null) || toColumn(n); // should a falsey header value be ignored? or just nullish?
     while (names.has(name)) name += "_";
     names.add(name);
   }
   names = new Array(c0).concat(Array.from(names));
 
   const output = new Array(r1 - r0 + 1);
   for (let r = r0; r <= r1; r++) {
-    const row = (output[r - r0] = Object.defineProperty({}, "#", {
+    // Should we be using Object.create(null) instead of an empty object here?
```
visnup, line 37: `Object.create(null)` should be better.

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
+    const row = (output[r - r0] = Object.defineProperty({}, "#", { // what is this non-enumerable row["#"] property for?
       value: r + 1,
     }));
-    const _row = sheet._rows[r];
+    const _row = sheet._rows[r]; // is this an internal ExcelJS API? why not sheet.getRow(r)?
```
visnup, line 41: Optimization? Can change to `getRow`, which means a row is always returned (if it was previously missing getRow creates one) and the check for `_row &&` below would be unneeded.

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
     if (_row && _row.hasValues)
       for (let c = c0; c <= c1; c++) {
-        const value = valueOf(_row._cells[c]);
-        if (value != null) row[names[c + 1]] = value;
+        const value = valueOf(_row._cells[c]); // internal ExcelJS API?
```
visnup, line 44: Public API would be:
```suggestion
        const value = valueOf(_row.findCell(c + 1));
```

src/xlsx.js
```diff
@@ -21,27 +21,28 @@ export class Workbook {
...
+        if (value != null) row[names[c + 1]] = value; // what if the name is “__proto__” e.g.?
```
visnup, line 45: `Object.create(null)` should fix that, right?

visnup (commented):
Sorry, at some point I switched from immediate comments back to review comments so this is the last batch of replies.

I'll go ahead and add a commit to make some of these changes.

src/xlsx.js
visnup: That was the specific Ramona case, but there was also no good reason the text had to be rich text. Her example also came from Numbers → Excel.

src/xlsx.js
visnup: Oof, for tests I want to bring in exceljs as a devDependency then. The minimal mockWorkbook function won't cut it anymore.

visnup: I did this in #255 as another PR on top of this one as it brings in ExcelJS as a devDependency to use in tests.

src/xlsx.js
mbostock: I was thinking of the value `false`. I agree that if it’s the empty string (which is also falsey) we want to replace that with a non-empty name. So, shouldn’t we coerce to a string here? 

```suggestion
    const value = headerRow ? valueOf(headerRow._cells[n]) : null;
    let name = value && (value += "") || toColumn(n);
```

src/xlsx.js
```diff
@@ -21,26 +21,25 @@ export class Workbook {
...
-function extract(sheet, {range, headers = false} = {}) {
+function extract(sheet, {range, headers} = {}) {
   let [[c0, r0], [c1, r1]] = parseRange(range, sheet);
-  const headerRow = headers && sheet._rows[r0++];
+  const headerRow = headers ? sheet._rows[r0++] : null;
   let names = new Set(["#"]);
   for (let n = c0; n <= c1; n++) {
-    let name = (headerRow ? valueOf(headerRow._cells[n]) : null) || toColumn(n);
+    const value = headerRow ? valueOf(headerRow._cells[n]) : null;
+    let name = value && (value += "") || toColumn(n);
```
mbostock, line 30: Oops, I meant:

```suggestion
    let name = value && (value + "") || toColumn(n);
```

visnup: Yay tests!

mbostock (commented):
LGTM. 👍

visnup (approved):

visnup:
Going to merge now and deal with all the version bumping and publishing tomorrow morning.

PR #255, opened 2021-09-15
Prefer public ExcelJS APIs

visnup:
Is ExcelJS as a devDependency ok? There was a previous risk in using a `mockWorkbook` function in tests previously and this guarantees better compatibility. But it's a whole new library to bring in.


mbostock (approved):
👍

Issue #231, opened 2021-07-28
Add options to FileAttachment(…).image()


Fil:
In https://github.com/observablehq/stdlib/blob/main/src/fileAttachment.js#L43 it would be interesting to add options such as:

* width
* height
* alt
* longdesc

A short-hand notation for "half-size" would also be useful (maybe `{pixelRatio: 2}`?) currently, when you upload a screenshot taken on a retina screen, you need to look up the image's width, then write a full `<img src="….url()" width="W/2">` tag.

mootari:
`longdesc` is deprecated and afaik not used anywhere. You probably meant `title`?

Fil:
oops, correct! Now I refreshed my memory by reading https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img and it seems that intrisicsize could also be useful in some cases. And also styles.

CobusT:
Also mentioned in the feedback repo: https://github.com/observablehq/feedback/issues/268

PR #267, opened 2021-10-18
add attributes to image() as options

CobusT:
Reference #231

This doesn't address all of the image options, but starts by addressing the most common requests... width, height, alt and title. 

Example:

Screen Shot 2021-10-17 at 5 12 23 PM

### DIFF
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -46,10 +46,10 @@ export class AbstractFile {
     const url = await this.url();
     return new Promise((resolve, reject) => {
       const i = new Image();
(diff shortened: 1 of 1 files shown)


src/fileAttachment.js
```diff
@@ -42,10 +42,10 @@ export class AbstractFile {
   async tsv(options) {
     return dsv(this, "\t", options);
   }
-  async image() {
+  async image({width, height}={}) {
```
mbostock, line 45:
```suggestion
  async image({width, height} = {}) {
```

PR #273, opened 2021-12-20
colorSpace: "display-p3"

Fil:
The third argument to DOM.context2d can be an options object with {scale, colorSpace}, allowing the creation of wide-gamut graphics on supported hardware and software.

Note: this necessitates a try/catch because Safari will throw if the os requirements are not met.

Tests/demo: https://observablehq.com/@fil/colorspace-display-p3-context2d

Draft, mostly for discussion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,9 +41,9 @@ html`<canvas width=960 height=500>`
 
 If you are using [2D Canvas](https://www.w3.org/TR/2dcontext/) (rather than [WebGL](https://webglfundamentals.org/)), you should use [DOM.context2d](#DOM_context2d) instead of DOM.canvas for automatic pixel density scaling.
 
-<a href="#DOM_context2d" name="DOM_context2d">#</a> DOM.<b>context2d</b>(<i>width</i>, <i>height</i>[, <i>dpi</i>]) [<>](https://github.com/observablehq/stdlib/blob/master/src/dom/context2d.js "Source")
+<a href="#DOM_context2d" name="DOM_context2d">#</a> DOM.<b>context2d</b>(<i>width</i>, <i>height</i>[, <i>options</i>]) [<>](https://github.com/observablehq/stdlib/blob/master/src/dom/context2d.js "Source")
 
-Returns a new canvas context with the specified *width* and *height* and the specified device pixel ratio *dpi*. If *dpi* is not specified, it defaults to [*window*.devicePixelRatio](https://developer.mozilla.org/docs/Web/API/Window/devicePixelRatio). To access the context’s canvas, use [*context*.canvas](https://developer.mozilla.org/docs/Web/API/CanvasRenderingContext2D/canvas). For example, to create a 960×500 canvas:
+Returns a new canvas context with the specified *width* and *height* and the specified *options*. If the options are specified as an object, its scale and colorSpace properties are considered. A string or number value specifies the scale. The scale defaults to [*window*.devicePixelRatio](https://developer.mozilla.org/docs/Web/API/Window/devicePixelRatio). The colorSpace defaults to srgb—the other acceptable value is display-p3. To access the context’s canvas, use [*context*.canvas](https://developer.mozilla.org/docs/Web/API/CanvasRenderingContext2D/canvas). For example, to create a 960×500 canvas:
 
 ```js
 {
--- a/src/dom/context2d.js
+++ b/src/dom/context2d.js
@@ -1,10 +1,22 @@
-export default function(width, height, dpi) {
-  if (dpi == null) dpi = devicePixelRatio;
+export default function(width, height, options) {
+  var colorSpace = "srgb";
+  var scale = devicePixelRatio;
+  if (options == +options) {
+    scale = +options;
+  } else {
+    if (options.scale) scale = +options.scale;
+    if (options.colorSpace === "display-p3") colorSpace = options.colorSpace;
+  }
   var canvas = document.createElement("canvas");
-  canvas.width = width * dpi;
-  canvas.height = height * dpi;
+  canvas.width = width * scale;
+  canvas.height = height * scale;
   canvas.style.width = width + "px";
-  var context = canvas.getContext("2d");
-  context.scale(dpi, dpi);
+  var context;
+  try {
+    context = canvas.getContext("2d", {colorSpace});
+  } catch(e) {
+    context = canvas.getContext("2d");
+  }
+  context.scale(scale, scale);
   return context;
 }


mbostock (changes requested):
We shouldn’t code for specific options and should instead support passthrough. If _options_ is an object, we pass it through to canvas.getContext, and we don’t add any try-catch around it. For backwards compatibility, if _options_ is a number, it’s promoted to an options object {scale}.

Although, I’m also tempted to do nothing here. For the most part the DOM API in standard library is deprecated. It’s better to just have people write against the standard APIs so that when code is copy-pasted outside of Observable there are no gratuitous dependencies on the standard library.

Fil:
It's fine to just do nothing. Easy enough to make this helper outside of stlib for special cases. (But it's very useful in general to not have to worry about dpr.)

src/dom/context2d.js
```diff
@@ -1,22 +1,20 @@
 export default function(width, height, options) {
-  var colorSpace = "srgb";
   var scale = devicePixelRatio;
-  if (options == +options) {
+  if (options == null) {
+    options = {};
```
mbostock, line 4: It’d be better to be strict here and only promote _options_ to an object when it is undefined, ideally using default arguments (`options = {}`). If the user passes in null explicitly, we should respect that and pass null to canvas.getContext.

src/dom/context2d.js
```diff
@@ -1,22 +1,20 @@
...
+  } else if (options == +options) {
```
mbostock, line 5: This should be strict based on the type (`typeof options === "number"`) rather than using coercion.

src/dom/context2d.js
```diff
@@ -1,22 +1,20 @@
...
     scale = +options;
+    options = {};
```
mbostock, line 7: This is probably fine, but it’d be slightly better to pass undefined, I think.

src/dom/context2d.js
```diff
@@ -1,22 +1,20 @@
...
   } else {
     if (options.scale) scale = +options.scale;
-    if (options.colorSpace === "display-p3") colorSpace = options.colorSpace;
+    delete options.scale;
```
mbostock, line 10: Oops, this mutates the user’s passed-in object! You’ll want to use the rest operator instead.

```js
const {scale, ...rest} = options;
options = rest;
```

mbostock (changes requested):

src/dom/context2d.js
Fil: Trying to be fully compatible with the current implementation, I wanted to support dpr passed as a string. Not sure why anyone would do that, though :)

mbostock: Full compatibility is not practical; someone could pass an object that implements valueOf, and this would have worked with the old API, but is ambiguous with named options under the new API. We should only support the intended pattern of passing a number.

Fil:
we might want to pass the [willReadFrequently](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/getContext) boolean option as well?

mootari:
I believe the options handling can be simplified:
```javascript
export function(width, height, options = {}) {
  const {
    scale = devicePixelRatio,
    ...contextOptions
  } = !isNaN(options) ? {scale: options} : options;
  const canvas = document.createElement("canvas");
  canvas.width = width * scale;
  canvas.height = height * scale;
  canvas.style.width = width + "px";
  const context = canvas.getContext("2d", contextOptions);
  context.scale(scale, scale);
  return context;
}
```

src/dom/context2d.mjs
```diff
@@ -1,10 +1,12 @@
-export default function(width, height, dpi) {
-  if (dpi == null) dpi = devicePixelRatio;
-  var canvas = document.createElement("canvas");
-  canvas.width = width * dpi;
-  canvas.height = height * dpi;
+export default function (width, height, options = {}) {
+  const {scale = devicePixelRatio, ...contextOptions} = !isNaN(options)
+    ? {...(options != null && {scale: options})}
+    : options;
```
mootari, line 4:
```suggestion
  const {scale = devicePixelRatio, ...contextOptions} = options && !isNaN(options)
    ? {scale: parseFloat(options)}
    : options;
```

src/dom/context2d.mjs
```diff
@@ -1,10 +1,12 @@
...
+  const canvas = document.createElement("canvas");
+  canvas.width = width * scale;
+  canvas.height = height * scale;
   canvas.style.width = width + "px";
-  var context = canvas.getContext("2d");
-  context.scale(dpi, dpi);
+  const context = canvas.getContext("2d", !options ? options : contextOptions);
```
mootari, line 9:
```suggestion
  const context = canvas.getContext("2d", contextOptions);
```

mootari (changes requested):

src/dom/context2d.mjs
Fil: This is meant to pass an explicit _null_ to the constructor, as per https://github.com/observablehq/stdlib/pull/273#discussion_r772564108 ; if we drop that requirement then the simpler code works for me.

mootari: In that case you'll want to do an explicit check for null: `options === null`

Fil: done

src/dom/context2d.mjs
Fil: 0 is currently working as before (in the sense that it returns a canvas of size 0)—it would be simpler if we didn't support this "feature".

README.md
```diff
@@ -41,9 +41,9 @@ html`<canvas width=960 height=500>`
 
 If you are using [2D Canvas](https://www.w3.org/TR/2dcontext/) (rather than [WebGL](https://webglfundamentals.org/)), you should use [DOM.context2d](#DOM_context2d) instead of DOM.canvas for automatic pixel density scaling.
 
-<a href="#DOM_context2d" name="DOM_context2d">#</a> DOM.<b>context2d</b>(<i>width</i>, <i>height</i>[, <i>dpi</i>]) [<>](https://github.com/observablehq/stdlib/blob/main/src/dom/context2d.mjs "Source")
+<a href="#DOM_context2d" name="DOM_context2d">#</a> DOM.<b>context2d</b>(<i>width</i>, <i>height</i>[, <i>options</i>]) [<>](https://github.com/observablehq/stdlib/blob/main/src/dom/context2d.mjs "Source")
```
mootari, line 44: Shouldn't we list both signatures here?

Fil: It's there: "As a shorthand notation, *options* having only scale can be specified as a number." (could be phrased better?)

src/dom/context2d.mjs
```diff
@@ -1,10 +1,15 @@
-export default function(width, height, dpi) {
-  if (dpi == null) dpi = devicePixelRatio;
-  var canvas = document.createElement("canvas");
-  canvas.width = width * dpi;
-  canvas.height = height * dpi;
+export default function (width, height, options = {}) {
+  const {scale = devicePixelRatio, ...contextOptions} = !isNaN(options)
+    ? {...(options != null && {scale: options})}
```
mootari, line 3: There's no need for the destructuring here, and the null check should be explicit. If it is undefined, we already get `{}`.
```suggestion
    ? {scale: options} ?? {}
```

src/dom/context2d.mjs
mootari: With the additional null check in the earlier comment we can remove `options &&` here.

README.md
mootari: I feel the shorthand should be mentioned earlier and emphasized, since it's the format that you'll encounter throughout Observable and the notation that you'll commonly use to disable dpr scaling.

Fil:
Chrome (>= 111) supports display-p3
https://developer.chrome.com/articles/high-definition-css-color-guide/

PR #283, opened 2022-04-26
SQLiteDatabaseClient schema methods

mbostock:
This implements the **describeTables** and **describeColumns** methods for SQLite.

Fixes #288.

### DIFF
--- a/src/sqlite.js
+++ b/src/sqlite.js
@@ -28,10 +28,16 @@ export class SQLiteDatabaseClient {
       text(rows.map(row => row.detail).join("\n"))
     ]);
   }
-  async describe(object) {
-    const rows = await (object === undefined
-      ? this.query(`SELECT name FROM sqlite_master WHERE type = 'table'`)
-      : this.query(`SELECT * FROM pragma_table_info(?)`, [object]));
+  async describeTables() {
+    return this.query(`SELECT name FROM sqlite_master WHERE type = 'table'`);
+  }
+  async describeColumns({table} = {}) {
+    const rows = await this.query(`SELECT name, type, notnull FROM pragma_table_info(?) ORDER BY cid`, [table]);
+    if (!rows.length) throw new Error(`table not found: ${table}`);
+    return rows.map(({name, type, notnull}) => ({name, type: sqliteType(type), databaseType: type, nullable: !notnull}));
+  }
+  async describe(table) {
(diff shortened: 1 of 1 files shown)


wiltsecarpenter (approved):

mbostock:
Asking @visnup for another opinion here on how to pass through the raw database type.

visnup:
I like this plus a single key called maybe `databaseType` or `vendorType` that includes the vendor-specific name for the type. And I'd want to re-work the schema object coming from data-connector to match this too, replacing the `int32: true`, `float32: true` entries with something like `databaseType: "int32"`

The only other thing worth considering is calling out bigint and date since the original [JSON schema specification on ajv](https://ajv.js.org/json-schema.html) didn't provide for either. These were `{type: "string", [bigint|date]: true}` since over the wire these values were serialized as strings.

It seems since then there's [JSON Type Definition](https://ajv.js.org/json-type-definition.html) which seems more promising?? And this [more recent JSON Schema reference](https://json-schema.org/understanding-json-schema/reference/string.html#built-in-formats) seems to document types for datetime.

mbostock:
I’ve populated the `databaseType` field. Here’s what else I would like to do, but does not appear possible with our SQLite implementation, sql.js:

* Return a query result schema on db.query. The `exec` function doesn’t return any schema information other than the column names, so we don’t know what the types of the returned columns are. We’re populating the **columns** field but that’s it.
* Coerce date strings to Date instances. SQLite has DATE and DATETIME types, but the returned query results for these types are strings. Since we lack type information on the query results (previous point), we can’t know which columns are dates and then coerce them. We could map the DATE and DATETIME types to strings on describe, but… I dunno if that’s better.

We could probably fork sql.js to fix these issues.

annie:
since this is a Table cell blocker, i think we should merge this once we've finalized the DatabaseClient schema API (based on [Wiltse's PR](https://github.com/observablehq/observablehq/pull/6682)) and updated this PR to match that specification (i've got a [draft PR](https://github.com/observablehq/stdlib/pull/288) in the wings to make those updates).

it seems reasonable to me to address the query results schema and date string issues as fast follows, if that sounds good? i can create separate issues for those!

annie (approved):

mbostock:
FYI, the describeColumns query had a syntax error which I fixed in d381b56a159498f84e1e4fe61c397eb7d2732878.

PR #285, opened 2022-05-04
add leaflet to recommended libraries

mbostock:
Adds [Leaflet](https://leafletjs.com/) to the list of recommended libraries (so that it’s easier for us to make a snippet for creating Leaflet maps).

### DIFF
--- a/bin/resolve-dependencies
+++ b/bin/resolve-dependencies
@@ -82,6 +82,10 @@ const mains = ["unpkg", "jsdelivr", "browser", "main"];
     const package = await resolve("mermaid");
     console.log(`export const mermaid = dependency("${package.name}", "${package.version}", "${package.export.replace(/\.core\.js$/, ".min.js")}");`);
   }
+  {
+    const package = await resolve("leaflet");
+    console.log(`export const leaflet = dependency("${package.name}", "${package.version}", "${package.export.replace(/-src\.js$/, ".js")}");`);
+  }
 })();
 
 async function resolve(specifier) {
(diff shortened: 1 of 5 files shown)


shancarter:
The default `L` construction feels a little unfortunate because it's so arcane in the context of a snippet in observable. It doesn't immediately identify itself as using leaflet, which is especially obscure for these magic std lib cars. Should we use `Leaflet` instead (like `Plot`) even though it's not the default global for leaflet?

shancarter:
I think this is a very useful snippet though, it seems like something I see people trying and struggling with quite a bit.

mbostock:
I don’t like the `L` either, but given that it’s the Leaflet convention (and Leaflet defines a global `L` which is used by some plugins), I think we should stick with that. In the autocomplete menu we can say “L [Leaflet]” like we do with Apache Arrow and Arquero.

> I think this is a very useful snippet though, it seems like something I see people trying and struggling with quite a bit.

Yes, people may struggle, but this is as simple as I can make it and it should work out of the box, which I think will encourage users to learn more and experience some success. Just think about how hard it is to do it right now without this snippet and without it being built-in to the standard library! 😄

mbostock:
Specifically I want to use `L` because if you look on the web for snippets of code using Leaflet, they will all use `L`, so users won’t have to translate the `L` into `Leaflet` to get them to work on Observable. E.g., https://leafletjs.com/examples/quick-start/

shancarter:
Sounds good. Having the description in the autocomplete will definitely help. Wonder if we could have the description in the title hover tag or whatever when we do them for standard lib objects?

shancarter (approved):
LGTM

PR #286, opened 2022-05-12
flare, miserables

mbostock:


tophtucker (approved):
k so flare's a hierarchy as csv, and miserables is a network as json. i guess we may want a json hierarchy and a csv network at some point, but if you think these are the more natural formats i trust you!

PR #289, opened 2022-06-02
allow Library.require override

mbostock:
This PR allows the user to change the require implementation by assigning to Library.require.

Originally, I intended this to be local to the Library instance (hence the _resolve_ argument to the Library constructor). However, there are a number of places where we want to require stuff statically, such as within the AbstractFile class definition. Therefore it seems unavoidable to have only a single definition of require. (Eventually we want to move everything to dynamic ES module imports anyway…)

### DIFF
--- a/README.md
+++ b/README.md
@@ -849,7 +849,7 @@ tex.options({displayMode: true})`E = mc^2`
 
 <a href="#require" name="require">#</a> <b>require</b>(<i>names…</i>) [<>](https://github.com/d3/d3-require/blob/master/index.js "Source")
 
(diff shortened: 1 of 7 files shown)


mbostock:
> Does this create the possibility of a race-condition if there are in-flight requires when setDefaultRequire is called?

You can’t really have multiple requires at the same time (because require is built around a single global function named define), so the expectation is that you set Library.require before you do anything else.

mbostock:
This was totally broken… sorry for not testing before asking you to review. I fixed this in 6de65c31f2711cdaa105cfc7fd18f4c246805950. It’d be nice to have a better way to test this locally via unit tests.

mbostock:
Also I needed to re-export requireFrom and resolveFrom in 784dcb3aca40a4c79d4cbb7cb7223630da38a7d8 so I could use them (without having to load d3-require twice, since d3-require is baked-in to this library).

PR #291, opened 2022-06-14
treat main as the default schema

mbostock:
Per https://sqlite.org/lang_attach.html

> The schema-names 'main' and 'temp' refer to the main database and the database used for temporary tables. The main and temp databases cannot be attached or detached. … If the name of the table is unique across all attached databases and the main and temp databases, then the schema-name prefix is not required. If two or more tables in different databases have the same name and the schema-name prefix is not used on a table reference, then the table chosen is the one in the database that was least recently attached.

### DIFF
--- a/src/sqlite.js
+++ b/src/sqlite.js
@@ -29,7 +29,7 @@ export class SQLiteDatabaseClient {
     ]);
   }
   async describeTables({schema} = {}) {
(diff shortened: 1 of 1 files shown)


annie (approved):
ooh `NULLIF` is nice

PR #293, opened 2022-06-23
python literals via pyodide

mbostock:


mootari:
Should the `pyodide` instance be exposed so that you can e.g. load `micropip`?

mbostock:
> Should the pyodide instance be exposed so that you can e.g. load micropip?

I’m not going to do this yet, but we might add this in the future.

mbostock:
Putting this on hold as, at a minimum, I think we need to update the [Observable Inspector](https://github.com/observablehq/inspector) to do a better job with PyProxy objects.

PR #295, opened 2022-06-28
mocha

mbostock:
Adopts Mocha for testing, which required renaming all .js to .mjs (because we’re not quite ready to adopt `"type": "module"` in our package.json yet).


annie (approved):
🙏

PR #296, opened 2022-07-01
Move __query and its dependencies to stdlib

annie:
This PR moves the __query function and its dependencies out of the Next worker code and into stdlib. I also moved a set of tests over for the query template function (which I renamed to `makeQueryTemplate`, which I find a bit clearer). 

Once we upgrade stdlib in Classic, this will serve as a partial fix for Data Table embeds (embeds will show the array of results, but not the full table UI).

### DIFF
--- a/src/table.mjs
+++ b/src/table.mjs
@@ -73,7 +73,7 @@ async function* accumulateQuery(queryRequest) {
  * of sub-strings and params are the parameter values to be inserted between each
  * sub-string.
  */
- function makeQueryTemplate(operations, source) {
+ export function makeQueryTemplate(operations, source) {
   const escaper =
     source && typeof source.escape === "function" ? source.escape : (i) => i;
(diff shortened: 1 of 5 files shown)


mbostock (approved):

src/table.mjs
```diff
@@ -0,0 +1,233 @@
...
+    yield values;
+  }
+}
+
+/**
+ * Returns a SQL query in the form [[parts], ...params] where parts is an array
+ * of sub-strings and params are the parameter values to be inserted between each
+ * sub-string.
+ */
+ export function makeQueryTemplate(operations, source) {
```
annie, line 76: we still need this function in Next because we use it to construct the query for the convert to SQL operation. should we also export it as part of the standard library? that feels wrong somehow ... maybe there's another way to share the code?

mbostock: Yes we can have it as an export in src/index.mjs

PR #301, opened 2022-09-08
dsv type option

mbostock:
Screen Shot 2022-09-07 at 6 40 32 PM

Open questions (with my inclination):

* Should we pass-through columns that are not mentioned without typing? (Yes.)
* Should we support the “drop” type, if we want to suppress a column? (Yes if pass-through is default.)
* Should we support the “auto” type, with the same behavior as d3.autoType? (Maybe.)
* Should we require ISO 8601 for dates rather than using the lax Date constructor? (Probably.)
* Any special handling for the empty string? (Probably not.)
* Should we use the Function constructor to make this faster, like d3-dsv does? (Maybe.)
* Or should we use Object.fromEntries instead? (Probably slower.)

### DIFF
--- a/src/fileAttachment.mjs
+++ b/src/fileAttachment.mjs
@@ -10,11 +10,38 @@ async function remote_fetch(file) {
   return response;
 }
 
-async function dsv(file, delimiter, {array = false, typed = false} = {}) {
+function asTyper(type) {
+  if (typeof type === "function") {
+    return type;
+  }
+  switch (`${type}`) {
+    case "number": return Number;
+    case "string": return String;
+    case "boolean": return Boolean;
+    case "date": return NewDate;
+    default: throw new Error(`unknown type: ${type}`);
+  }
+}
+
+function NewDate(value) {
+  return new Date(value);
+}
+
+function asTypers(types) {
+  const typers = Object.create(null);
+  for (const key in types) typers[key] = asTyper(types[key]);
+  return (d, i) => {
+    const object = Object.create(null);
+    for (const key in typers) object[key] = typers[key](d[key], i);
+    return object;
+  };
+}
+
+async function dsv(file, delimiter, {array = false, type, typed = false} = {}) {
   const text = await file.text();
   return (delimiter === "\t"
       ? (array ? tsvParseRows : tsvParse)
-      : (array ? csvParseRows : csvParse))(text, typed && autoType);
+      : (array ? csvParseRows : csvParse))(text, type ? asTypers(type) : typed ? autoType : null);
 }
 
 export class AbstractFile {


src/fileAttachment.mjs
```diff
@@ -10,11 +10,38 @@ async function remote_fetch(file) {
...
 
-async function dsv(file, delimiter, {array = false, typed = false} = {}) {
+function asTyper(type) {
+  if (typeof type === "function") {
+    return type;
+  }
+  switch (`${type}`) {
+    case "number": return Number;
+    case "string": return String;
+    case "boolean": return Boolean;
```
enjalot, line 20: this differs from autotype in that it will not turn "false" into `false`. is that ok?

src/fileAttachment.mjs
```diff
@@ -10,11 +10,38 @@ async function remote_fetch(file) {
...
 }
 
-async function dsv(file, delimiter, {array = false, typed = false} = {}) {
+function asTyper(type) {
+  if (typeof type === "function") {
+    return type;
+  }
+  switch (`${type}`) {
+    case "number": return Number;
+    case "string": return String;
```
enjalot, line 19: in autoType the string gets trimmed. probably not something to include in casting, but will be different

src/fileAttachment.mjs
```diff
@@ -10,11 +10,38 @@ async function remote_fetch(file) {
...
+    case "boolean": return Boolean;
+    case "date": return NewDate;
+    default: throw new Error(`unknown type: ${type}`);
+  }
+}
+
+function NewDate(value) {
+  return new Date(value);
```
enjalot, line 27: do we want to do the timezone workaround for Safari?

enjalot:
> Should we pass-through columns that are not mentioned without typing? (Yes. At the end?)  

I agree that unmentioned columns be included by default

> Should we support the “drop” type, if we want to suppress a column? (Yes if pass-through is default.)

I agree, nice to explicitly drop unwanted columns (also documents the decision nicely)

> Should we support the “auto” type, with the same behavior as d3.autoType for a single column? (Maybe.)

I don't think so. The primary interface for this implies that auto already suggested the type for the column. I like this level of interface being explicit.

> Should we support the “bigint” type? (Probably.)

Seems like something we want to support across the platform better right?

> Should we require ISO 8601 for dates rather than using the lax Date constructor? (Probably.)

i agree, we should standardize. if someone wants a variation the inline function is easy to use.

> Any special handling for the empty string? (Probably not.)

I don't think so.

> Should we use the Function constructor to make this faster, like d3-dsv does? (Maybe.)
Or should we use Object.fromEntries? (Probably slower.)

Don't have a strong opinion, but making it as fast as possible seems a worthy goal.

src/fileAttachment.mjs
mbostock: Ah, no, that wasn’t intentional.

PR #304, opened 2022-10-07
Add __table function and unit tests

annie:
This PR ports over @mbostock's `__table` function, as defined in https://github.com/observablehq/observablehq/pull/8300, and adds missing support for the `lt` and `gt` filter operations. Also adds unit tests.

### DIFF
--- a/src/table.mjs
+++ b/src/table.mjs
@@ -1,5 +1,25 @@
 import {ascending, descending, reverse} from "d3-array";
 
+// We support two levels of DatabaseClient. The simplest DatabaseClient
+// implements only the client.sql tagged template literal. More advanced
+// DatabaseClients implement client.query and client.queryStream, which support
+// streaming and abort, and the client.queryTag tagged template literal is used
+// to translate the contents of a SQL cell or Table cell into the appropriate
+// arguments for calling client.query or client.queryStream. For table cells, we
+// additionally require client.describeColumns. The client.describeTables method is
+// optional.
+export function isDatabaseClient(value, mode) {
+  return (
+    value &&
+    (typeof value.sql === "function" ||
+      (typeof value.queryTag === "function" &&
+        (typeof value.query === "function" ||
+          typeof value.queryStream === "function"))) &&
+    (mode !== "table" || typeof value.describeColumns === "function") &&
+    value !== __query // don’t match our internal helper
+  );
+}
+
 export function isTypedArray(value) {
   return (
     value instanceof Int8Array ||
@@ -9,7 +29,8 @@ export function isTypedArray(value) {
     value instanceof Uint8ClampedArray ||
     value instanceof Uint16Array ||
     value instanceof Uint32Array ||
-    value instanceof Float32Array
+    value instanceof Float32Array ||
+    value instanceof Float64Array
   );
 }
 
@@ -18,10 +39,9 @@ export const __query = Object.assign(
   async (source, operations, invalidation) => {
     // For cells whose data source is an in-memory table, we use JavaScript to
     // apply the table cell operations, instead of composing a SQL query.
-    // TODO Do we need to port over all the logic from canDisplayTable, or is
-    // this sufficient?
     if (Array.isArray(source) || isTypedArray(source)) return __table(source, operations);
-    const args = makeQueryTemplate(await source, operations);
+    if (!isDatabaseClient(source)) throw new Error("invalid source");
+    const args = makeQueryTemplate(operations, await source);
     if (!args) return null; // the empty state
     return evaluateQuery(await source, args, invalidation);
   },
@@ -93,7 +113,7 @@ async function* accumulateQuery(queryRequest) {
  * of sub-strings and params are the parameter values to be inserted between each
  * sub-string.
  */
- export function makeQueryTemplate(source, operations) {
+ export function makeQueryTemplate(operations, source) {
   const escaper =
     source && typeof source.escape === "function" ? source.escape : (i) => i;
   const {select, from, filter, sort, slice} = operations;
--- a/src/table.mjs
+++ b/src/table.mjs
@@ -1,7 +1,14 @@
+import {ascending, descending, reverse} from "d3-array";
+
 export const __query = Object.assign(
   // This function is used by table cells.
   async (source, operations, invalidation) => {
-    const args = makeQueryTemplate(operations, await source);
+    // For cells whose data source is an in-memory table, we use JavaScript to
+    // apply the table cell operations, instead of composing a SQL query.
+    // TODO Do we need to do a more comprehensive isArray check, like the one
+    // we do in the worker?
+    if (Array.isArray(source)) return __table(source, operations);
+    const args = makeQueryTemplate(await source, operations);
     if (!args) return null; // the empty state
     return evaluateQuery(await source, args, invalidation);
   },
@@ -73,7 +80,7 @@ async function* accumulateQuery(queryRequest) {
  * of sub-strings and params are the parameter values to be inserted between each
  * sub-string.
  */
- export function makeQueryTemplate(operations, source) {
+ export function makeQueryTemplate(source, operations) {
   const escaper =
     source && typeof source.escape === "function" ? source.escape : (i) => i;
   const {select, from, filter, sort, slice} = operations;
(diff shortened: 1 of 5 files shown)


test/table-test.mjs
```diff
@@ -35,178 +35,261 @@ const baseOperations = {
...
-    }
-  ];
-
-  invalidFilters.map((filter) => {
-    const operations = {
-      ...baseOperations,
-      filter: [filter]
-    };
-    assert.throws(() => makeQueryTemplate(operations, source), /Invalid filter operation/);
+describe("makeQueryTemplate", () => {
```
annie, line 38: the only change i made here was to wrap the `makeQueryTemplate` tests in a `describe` – if you check "Hide whitespace", this diff will look much better :)

annie, line 38: the only change i made here was to wrap the `makeQueryTemplate` tests in a `describe` – if you check "Hide whitespace", this diff will look much better 😄

src/table.mjs
```diff
@@ -1,7 +1,27 @@
...
+  );
+}
+
 export const __query = Object.assign(
   // This function is used by table cells.
   async (source, operations, invalidation) => {
-    const args = makeQueryTemplate(operations, await source);
+    // For cells whose data source is an in-memory table, we use JavaScript to
+    // apply the table cell operations, instead of composing a SQL query.
+    // TODO Do we need to port over all the logic from canDisplayTable, or is
```
annie, line 21: [canDisplayTable](https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/display.js#L24) contains a lot of logic, and i don't know if we need to copy that over to stdlib? it seems like checking the general structure of `source` is sufficient here, since the source needs to have passed the `canDisplayTable` check to even show up as a data source option in the first place.

src/table.mjs
```diff
@@ -1,7 +1,27 @@
...
+    // this sufficient?
+    if (Array.isArray(source) || isTypedArray(source)) return __table(source, operations);
+    const args = makeQueryTemplate(await source, operations);
```
mbostock, line 24: At this point, we could throw an error if the source is not a database client. We have an isDatabaseClient function somewhere that does the duck test.

src/table.mjs
```diff
@@ -1,7 +1,27 @@
...
+export function isTypedArray(value) {
+  return (
+    value instanceof Int8Array ||
+    value instanceof Int16Array ||
+    value instanceof Int32Array ||
+    value instanceof Uint8Array ||
+    value instanceof Uint8ClampedArray ||
+    value instanceof Uint16Array ||
+    value instanceof Uint32Array ||
+    value instanceof Float32Array
```
mbostock, line 12: TODO Float64Array?

src/table.mjs
```diff
@@ -73,7 +93,7 @@ async function* accumulateQuery(queryRequest) {
  * of sub-strings and params are the parameter values to be inserted between each
  * sub-string.
  */
- export function makeQueryTemplate(operations, source) {
+ export function makeQueryTemplate(source, operations) {
```
annie, line 96: note to self: revert this change because `makeQueryTemplate` is exported, so making this change would require a major version bump

src/table.mjs
annie: done

src/table.mjs
annie: done 🤞

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+      }
+      default:
+        throw new Error(`unknown filter type: ${type}`);
+    }
+  }
+  for (const {column, direction} of reverse(operations.sort)) {
+    const compare = direction === "desc" ? descending : ascending;
+    source.sort((a, b) => compare(a[column], b[column]));
+  }
+  if (operations.slice) {
```
mbostock, line 341: According to our schema, operations.slice is always present, but we only need to slice if operations.slice.from or operations.slice.to is non-null (or non-zero and non-infinity, if preferred).

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+    source = source.slice(
+      operations.slice.from ?? 0,
+      operations.slice.to ?? Infinity
+    );
+  }
+  if (operations.select?.columns) {
```
mbostock, line 347: And likewise here, options.select is required, so we don’t need the `?`.

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+      schema = operations.select.columns.map((c) => schemaByName.get(c));
+    } else if (columns) {
+      columns = operations.select.columns;
+    }
+    source = source.map((d) =>
+      Object.fromEntries(operations.select.columns.map((c) => [c, d[c]]))
+    );
+  }
+  if (schema) source.schema = schema;
+  else if (columns) source.columns = columns;
```
mbostock, line 359: We should move this assignment to within the if (operations.select.columns) block; otherwise, we should allow the source’s existing schema and columns to pass through unaffected, and we shouldn’t mutate the passed-in source (i.e., we only want to assign these properties after creating a copy of source by calling source.map).

src/table.mjs
```diff
@@ -1,6 +1,46 @@
...
+    (typeof value.sql === "function" ||
+      (typeof value.queryTag === "function" &&
+        (typeof value.query === "function" ||
+          typeof value.queryStream === "function"))) &&
+    (mode !== "table" || typeof value.describeColumns === "function") &&
+    value !== __query // don’t match our internal helper
+  );
+}
+
+export function isTypedArray(value) {
```
mbostock, line 23: This does not need to be exported.

```suggestion
function isTypedArray(value) {
```

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
   return {...operand, value: `%${operand.value}%`};
 }
 
+// This function applies table cell operations to an in-memory table (array of objects).
+export function __table(source, operations) {
+  let {schema, columns} = source;
+  for (const {type, operands} of operations.filter) {
+    const column = operands.find(({type}) => type === "column").value;
```
mbostock, line 278: According to our schema, the column operand must always be listed first. So, in theory we could write it faster like this:

```suggestion
    const [{value: column}] = operands;
```

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+    const resolved = operands.filter(({type}) => type === "resolved");
```
mbostock, line 279: I’m not sure this is right. According to our schema, TableFilterOperands can have the following type:

* _primitive_
* _date_
* _cell_

I don’t see us using the type _resolved_ anywhere. I’m also not sure what the type _cell_ means… I’m guessing that was speculative and should be removed from the schema. Anyway, I think we can say this:

```suggestion
    const values = operands.slice(1).map(({value}) => value);
```

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+        source = source.filter((d) => d[column] > value);
+        break;
+      }
+      default:
+        throw new Error(`unknown filter type: ${type}`);
+    }
+  }
+  for (const {column, direction} of reverse(operations.sort)) {
+    const compare = direction === "desc" ? descending : ascending;
+    source.sort((a, b) => compare(a[column], b[column]));
```
mbostock, line 339: This might be mutating the input _source_. We need to make sure we make a defensive copy before sorting (but ideally we avoid unnecessary defensive copies, too).

mbostock (changes requested):
If I have time tonight I’ll put up a PR to this PR to help this along! Thanks Annie!

src/table.mjs
mbostock: Ah ha. The compiler replaces _primitive_, _date_, and _cell_ filter operands with _resolved_. So it was correct as-is. I’ll update my PR accordingly.

src/table.mjs
```diff
@@ -231,3 +271,91 @@ function likeOperand(operand) {
...
+// This function applies table cell operations to an in-memory table (array of objects).
+export function __table(source, operations) {
+  let {schema, columns} = source;
+  for (const {type, operands} of operations.filter) {
+    const column = operands.find(({type}) => type === "column").value;
+    const resolved = operands.filter(({type}) => type === "resolved");
+    switch (type) {
+      case "eq": {
+        const [{value}] = resolved;
+        source = source.filter((d) => d[column] === value);
```
mbostock, line 283: We’re using strict equality here, but the resolved value will always be a string (in our UI), whereas the type of the column might be a number. I think this means we’ll have to use loose equality here instead… and I’m not sure how we’ll handle dates.

libbey-observable (approved):

mbostock (approved):

src/table.mjs
```diff
@@ -1,12 +1,50 @@
...
 export const __query = Object.assign(
-  // This function is used by table cells.
   async (source, operations, invalidation) => {
-    const args = makeQueryTemplate(operations, await source);
-    if (!args) return null; // the empty state
-    return evaluateQuery(await source, args, invalidation);
+    source = await source;
+    // For cells whose data source is an in-memory table, we use JavaScript to
+    // apply the table cell operations, instead of composing a SQL query.
+    if (Array.isArray(source) || isTypedArray(source)) return __table(source, operations);
```
mbostock, line 43: The isTypedArray check here isn’t enough—we also need to promote the typed array into an array of {value} objects so that __table can work as expected.

src/table.mjs
mbostock: I’ve ported this logic over so that we can show the correct errors and also handle promotion of primitive arrays.

PR #310, opened 2022-10-23
Add DuckDBClient as a recommended library

### LINKED ISSUE #308: Add DuckDBClient to recommended libraries
Using [DuckDB](https://duckdb.org/), users can create a database from `.csv`, `.json`, or `.parquet` files in a single line of code. Then, they can use familiar sql syntax (in a sql cell) to query their data. 

This is an excellent entry point for data analysts unfamiliar with JavaScript, and has been repeatedly used by for demos and tutorials (see [EDA workshop](https://observablehq.com/@observablehq/eda-workshop-materials), [PSI workshop](https://observablehq.com/@observablehq/psi-workshop)). 

This is especially important as it allows us to load data without introducing the construct of imports, which could feel overwhelming for new users.

mkfreeman:
Resolves #308

Outstanding issues: 
- I was unable to use the `dependency(...)` pattern because the default file of `"@duckdb/duckdb-wasm"` is a `.cjs` file: `dist/duckdb-browser.cjs` (this then causes the build to fail)
- Requires upgrading to apache arrow version 8
- Now exposes the entire DuckDBClient ~~The current implementation only exposes the `.of` method of the [DuckDBClient](https://observablehq.com/@cmudig/duckdb-client), which is really opinionated. But, this allows the really concise and readable syntax `db = duckdb(FileAttachment("traffic.csv"))`, or `duckdb({penguins})`, which feels intuitive to me~~


mbostock (approved):

domoritz:
This is great. Thanks for putting this together. Any recommendations for how I can encourage people to get off of the client from https://observablehq.com/@cmudig/duckdb when this pull request is merged? I can add a message at the top but was wondering whether you have some recommendations beyond that.

mbostock:
@domoritz Thanks! I’ll put up a notebook once its released and yes, it’d be great if you could add a banner at the top of your notebook that directs people to Observable’s new DuckDBClient. I don’t have any other recommendations at this time, but I’m open to suggestions. Maybe it’d be nice if you could notify people importing your notebook somehow…

domoritz:
Before you publish the library, would you mind making a pre-release that I can try in a notebook just to see whether there are any issues I find that way?

mbostock:
Here, try this:

```js
import {DuckDBClient} from "1710adacec0d31ba"
```

https://observablehq.com/d/1710adacec0d31ba

domoritz:
```js
c = new DuckDBClient()
c.query(`SELECT
  v::INT AS x,
  (sin(v/50.0) * 100 + 100)::INT AS y
FROM generate_series(0, 1000) AS t(v)`)
```

doesn't work anymore (`TypeError: undefined is not an object (evaluating 'this._db.connect')`). Would it make sense to init the db in the constructor?

domoritz:
Yep, that works. I wonder whether it would still be good to just init a db if you don't get one in the constructor. Ignore my suggestion if the common pattern for observable DBs is to not do that.

mbostock:
> I wonder whether it would still be good to just init a db if you don't get one in the constructor.

It requires loading duckdb-wasm which is async, and constructors have to be synchronous.

PR #312, opened 2022-11-01
MS SQL syntax for LIMIT/OFFSET

Sylvestre67:
Resolves https://github.com/observablehq/projects/issues/269.

SQL server syntax for LIMIT/OFFSET type query is
```SQL
ORDER BY <col> <direction>
OFFSET 0 ROWS
FETCH NEXT 1000 ROWS ONLY
```

The table cell is currently generating the standard SQL syntax only.
```SQL
LIMIT 1000
OFFSET 0
```
If the source is of dialect `mssql` it should use the right `LIMIT/OFFSET` syntax.

MS SQL [documentation](https://learn.microsoft.com/en-us/sql/t-sql/queries/select-order-by-clause-transact-sql?view=sql-server-ver16#using-offset-and-fetch-to-limit-the-rows-returned).

### DIFF
--- a/src/table.mjs
+++ b/src/table.mjs
@@ -256,7 +256,6 @@ export function makeQueryTemplate(operations, source) {
   }
   if (source.dialect === "mssql") {
     if (slice.to !== null || slice.from !== null) {
-
       if (!sort.length) {
         if (!select.columns)
           throw new Error(


mbostock (commented):
Do you mind running Prettier on the parts you’ve changed? I see some formatting nits here and I’d rather not comment on them. Thanks!

src/table.mjs
```diff
@@ -254,14 +254,28 @@ export function makeQueryTemplate(operations, source) {
...
-      args
-    );
-  }
-  if (slice.from !== null) {
-    appendSql(` OFFSET ${slice.from}`, args);
+  if(source.dialect === 'mssql'){
+    if (slice.to !== null || slice.from !== null) {
+      if(!sort.length){
+        appendSql(`\nORDER BY `, args);
+        appendOrderBy({column: select.columns[0], direction: 'ASC'}, args);
```
mbostock, line 261: I think we’ll want the escaped column name here, to match the SELECT.

```suggestion
        appendOrderBy({column: columns[0], direction: 'ASC'}, args);
```

Also, this will crash if select.columns is null, as it is with `SELECT *`. In practice this should never happen for a MS SQL data source because we would always materialize the selected columns in a table cell, but we should either throw an error here with a human-readable message or perhaps we can make it work with something like ORDER BY NULL.

mbostock (changes requested):

src/table.mjs
```diff
@@ -248,20 +248,48 @@ export function makeQueryTemplate(operations, source) {
   ];
   for (let i = 0; i < filter.length; ++i) {
     appendSql(i ? `\nAND ` : `\nWHERE `, args);
+    if (!filter[i].operands) throw new Error("missing operands");
+    filter[i].operands.forEach((op) => {
+      if (op.type === "column") op.value = escaper(op.value);
```
mbostock, line 253: This is mutating the passed-in _operations_ object. I’d guess that’s not harmful under the current usage, but isn’t something we should do in general since in theory that operations object could be reused by the caller for something else. For example if you called makeQueryTemplate twice with the same _operations_ object, the column name would get double-escaped.

So instead of mutating _op_ here, we should do the escaping as needed when interpolating into a SQL string.

src/table.mjs
```diff
@@ -248,20 +248,48 @@ export function makeQueryTemplate(operations, source) {
...
+    });
     appendWhereEntry(filter[i], args);
   }
   for (let i = 0; i < sort.length; ++i) {
     appendSql(i ? `, ` : `\nORDER BY `, args);
+    sort[i].column = escaper(sort[i].column);
```
mbostock, line 259: Same here re. mutation.

mbostock (changes requested):
Good find on the need for more escaping! Thanks for adding tests, too. 🙏

mbostock:
Sent #314 for your consideration. 🙏

src/table.mjs
```diff
@@ -248,20 +248,44 @@ export function makeQueryTemplate(operations, source) {
...
-      `\nLIMIT ${slice.to !== null ? slice.to - (slice.from || 0) : 1e9}`,
-      args
-    );
-  }
-  if (slice.from !== null) {
-    appendSql(` OFFSET ${slice.from}`, args);
+  if (source.dialect === "mssql") {
+    if (slice.to !== null || slice.from !== null) {
+      if (!sort.length) {
+        if (columns[0] === "*")
```
mbostock, line 260: This isn’t how we represent SELECT *; when selecting all columns, `columns` will be null here (and so this will crash as currently written). See this test for example:

https://github.com/observablehq/stdlib/blob/f4a8b6de4a76d078063c8f4a0fe352aaf3781d3c/test/table-test.mjs#L50-L54

mbostock (changes requested):
Almost there…

src/table.mjs
mbostock: Oh, you’re right; I was confusing `select.columns` (from the passed-in `operations`) and the local variable `columns`. You could use `columns === "*"` I suppose, then. But I think `!select.columns` would be more explicit and preferable.

mbostock (approved):
👍👍

PR #313, opened 2022-11-02
DuckDBClient

mbostock:
This…

* Mostly rewrites CMU’s DuckDBClient (with a proper implementation of queryStream).
* Minimizes the exposed interface for DuckDBClient to reduce maintenance burden and discourage mutation.
* Preserves stdlib’s `Arrow` as apache-arrow@4 for backwards compatibility.
* Adds apache-arrow@9 (not 8) as a new dependency for DuckDB, per [its package.json](https://github.com/duckdb/duckdb-wasm/blob/58f1203f913f06340606ca9bee0b038360d90981/packages/duckdb-wasm/package.json#L26).
* Uses [jsDelivr’s esm.run](https://www.jsdelivr.com/esm) to load @duckdb/duckdb-wasm and apache-arrow as an ES module (rather than require).
* Minimizes the changes to recommended library versions in src/dependencies.mjs.
* Adds CMU DIG’s BSD license to the derived code for DuckDBClient.

Demo:

Screen Shot 2022-11-02 at 3 31 07 PM

TODO Document and test the DuckDBClient implementation.


domoritz:
Oh cool. I'll take a look.

kimmolinna:
@mbostock this sounds really good. 👍

src/duckdb.mjs
```diff
@@ -1,361 +1,228 @@
...
-      if (!this._db) {
-        this._db = await makeDB();
-        await this._db.open({
-          query: {
-            castTimestampToDate: true
+      })),
+      async *readRows() {
+        try {
+          while (!batch.done) {
+            yield batch.value.toArray();
```
mbostock, line 64: Not as part of this PR, but we should have an option to return Apache Arrow Table instances directly from the database client, rather than converting to an array of objects (rows) here. Probably this is an option (arrow = true) when you run a query, and perhaps you can specify it as a default option when constructing the DuckDBClient. In conjunction with #315 it would mean that we could entirely avoid materializing the array of objects when displaying the results of a DuckDBClient query in a table.

kimmolinna: At the moment `DuckDBClient.query` returns Apache Arrow Table so this sounds good once again.

kimmolinna: And It would be really great if you could use Apache Arrow Table also with Plot.

mbostock: @kimmolinna Agreed! Tracking that here: https://github.com/observablehq/plot/issues/1103

src/duckdb.mjs
```diff
@@ -1,361 +1,228 @@
...
   }
 
-  // Adapted from: https://observablehq.com/@cmudig/duckdb-client
-  // Follows the DatabaseClient specification: https://observablehq.com/@observablehq/database-client-specification
-  class DuckDBClient {
-    constructor(_db) {
-      this._db = _db;
-      this._counter = 0;
+  async queryStream(query, params) {
+    const connection = await this._db.connect();
```
mkfreeman, line 44: What type of object is `this._db` that has the `.connect` method?

mkfreeman: A database client? Not seeing a `.connect` method in the [specification](https://observablehq.com/@observablehq/database-client-specification)

src/duckdb.mjs
```diff
@@ -1,361 +1,228 @@
...
+    };
+  }
 
-      if (params) {
-        const stmt = await conn.prepare(query);
-        result = stmt.query(...params);
-      } else {
-        result = await conn.query(query);
+  async query(query, params) {
+    const result = await this.queryStream(query, params);
```
mkfreeman, line 75: Do we need to have an implementation of `.query` that doesn't rely on `queryStream` for some back-compatibility with existing DuckDB notebooks (e.g., [this cel](https://observablehq.com/@cmudig/duckdb-client#cell-16)l)? I imagine it's fine, as those imported versions of `DuckDBClient` will overwrite the standard library version, but just wanted to surface this as a possible concern.

mkfreeman: (leaving some notes as we're pairing on the review, thinking out loud a bit) - does this mean that all DuckDBClients support streaming (as this uses the `queryStream` method)?

src/duckdb.mjs
```diff
@@ -1,361 +1,228 @@
...
+    return results;
+  }
 
-    // Summarize a query result
-    async summarize(query) {
-      const result = await this.query(`SUMMARIZE ${query}`);
-      return Inputs.table(result);
-    }
+  async queryRow(query, params) {
+    const results = await this.query(query, params);
```
mkfreeman, line 87: Since `this.query` calls `this.queryStream` as the first line, should we just call `queryStream` directly?

src/duckdb.mjs
mbostock: No, it is an AsyncDuckDB instance created here:

https://github.com/observablehq/stdlib/blob/a0fef0c392cbfa67ada6a65d91ddad46f11b3280/src/duckdb.mjs#L120

The connect method is documented here:

https://shell.duckdb.org/docs/classes/index.AsyncDuckDB.html#connect

src/duckdb.mjs
```diff
@@ -1,361 +1,228 @@
...
-      await db.registerFileBuffer(name, new Uint8Array(buffer));
-      const conn = await db.connect();
-      await conn.insertJSONFromPath(name, {
-        name,
-        schema: "main",
-        ...options
-      });
-      await conn.close();
+  async sql(strings, ...args) {
+    return await this.query(strings.join("?"), args);
```
mkfreeman, line 92: Same comment as above (should we just use `queryStream`?)

src/duckdb.mjs
mbostock: Anything in the standard library can be overridden, so any notebook that defines its own DuckDBClient won’t see the implementation here.

As far as this implementation goes, the fact that it implements _db_.query on top of _db_.queryStream should be totally invisible to the user. It implements the _db_.query method as specified in the database client specification and it only calls _db_.queryStream under the hood to reduce code duplication. It does not somehow mean that _db_.query or _db_.queryRow supports streaming; those methods still just return a promise to the results.

I should also note that I didn’t implement the signal option for query methods in the DuckDBClient so there isn’t currently a way to cancel queries. But in practice that probably shouldn’t matter much because it’s an in-memory client and therefore most queries should be relatively fast. We could add it in the future.

src/duckdb.mjs
mbostock: We could do that, but then we’d have to duplicate the rest of _db_.query, too (to call _result_.readRows _e.g._). But I guess that’s worth it since we don’t need to read multiple batches just to get the first row.

src/duckdb.mjs
mbostock: No, then we’d need to duplicate the rest of _db_.query. _db_.queryStream doesn’t return a promise to an array of results; it returns a [query stream response](https://observablehq.com/@observablehq/database-client-specification#%C2%A73.2).

src/duckdb.mjs
mbostock: Optimized in 6bf7aea.

src/duckdb.mjs
mkfreeman: Got it, thanks for the clarification!

libbey-observable (approved):
Did some basic testing locally after pairing on the review with @mkfreeman. Looks good!

bin/resolve-dependencies
```diff
@@ -64,7 +64,11 @@ const mains = ["unpkg", "jsdelivr", "browser", "main"];
   }
   {
     const package = await resolve("apache-arrow@4");
-    console.log(`export const arrow = dependency("${package.name}", "${package.version}", "${package.export}");`);
+    console.log(`export const arrow4 = dependency("${package.name}", "${package.version}", "${package.export}");`);
+  }
+  {
+    const package = await resolve("apache-arrow@9");
+    console.log(`export const arrow9 = dependency("${package.name}", "${package.version}", "+esm");`);
```
domoritz, line 71: Arrow 10 is out so you can update to that.

src/duckdb.mjs
domoritz: Sweet. I think native arrow support throughout would be awesome since we reduce materializations.

src/duckdb.mjs
```diff
@@ -1,361 +1,234 @@
...
-        schema: "main",
-        ...options
-      });
-      await conn.close();
+async function insertArray(database, name, array, options) {
+  const arrow = await loadArrow();
+  const table = arrow.tableFromJSON(array);
+  const buffer = arrow.tableToIPC(table);
+  const connection = await database.connect();
+  try {
```
domoritz, line 196: I think it would be good to pull out a method for inserting an arrow `Table`. This method would be used to insert data that is already in columnar form.

src/duckdb.mjs
```diff
@@ -1,361 +1,234 @@
...
+          schema: "main",
+          ...options
+        });
+        break;
+      default:
+        if (file.name.endsWith(".parquet")) {
+          await connection.query(
+            `CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`
+          );
+        } else {
```
domoritz, line 182: Add a case for `file.name.endsWith(".arrow")`

src/duckdb.mjs
```diff
@@ -1,361 +1,234 @@
...
-      }
-      // Populate the schema of the results
-      const schema = result.schema.fields.map(({
-        name,
-        type
-      }) => ({
+    return {
+      schema: schema.fields.map(({name, type}) => ({
         name,
         type: getType(String(type)),
```
domoritz, line 58: This is maybe more brittle than directly checking against Arrow types but probably okay for now. Can improve later.

domoritz (approved):
This is great. Just added some comments for more flexible imports.

src/duckdb.mjs
mbostock: Yes, I can use the Apache Arrow type identifiers when looking at the schema for an Apache Arrow Table, but the result of a DESCRIBE doesn’t give me the numeric identifiers — I only get the string column_type. I want to use the same code for describeColumns, too. It’d be nice if DuckDB returned the numeric type identifier in addition to the string.

mbostock: Added in 990c240. And I was able to simplify the switch for DuckDB types since now we only need to handle the canonical types returned by DESCRIBE rather than the aliases. (At least, I hope that’s the case… it appears to be in my local testing.)

src/duckdb.mjs
mbostock: Added in 71fbda3.

src/duckdb.mjs
mbostock: Added in 71fbda3.

bin/resolve-dependencies
mbostock: Going to stick with Arrow 9 until duckdb-wasm upgrades…

https://github.com/duckdb/duckdb-wasm/blob/d779c7d5b6758a2abade699d1b564e1140a68de1/packages/duckdb-wasm/package.json#L26

PR #315, opened 2022-11-05
Apache Arrow for table cells

mbostock:
TODO

- [ ] Implement table operations in __arrow


mbostock:
Mostly folded into #313, leaving only the need to implement __table for Apache Arrow tables… which I’m not sure how to do, so closing for now.

PR #322, opened 2022-11-15
Instantiate DuckDBClient for Arrow tables

libbey-observable:
Works with [#8953](https://github.com/observablehq/observablehq/pull/8953) in observablehq

This is a small change to instantiate a DuckDBClient for Arrow tables, but it comes with a lot of questions:
- will it cause side effects to instantiate here?
- will we be instantiating a DuckDBClient too often?
- how expensive is it to instantiate one?
- is it possible to do it in a way that persists, instead?

Demo:  

arrow_in_table

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -147,7 +147,6 @@ export const __query = Object.assign(
     source = await loadDataSource(await source, "table");
     if (isDatabaseClient(source)) return evaluateQuery(source, makeQueryTemplate(operations, source), invalidation);
     if (isDataArray(source)) return __table(source, operations);
-    if (isArrowTable(source)) return __arrow(source, operations, invalidation);
     if (!source) throw new Error("missing data source");
     throw new Error("invalid data source");
   },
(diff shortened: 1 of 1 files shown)


src/table.js
```diff
@@ -145,6 +147,7 @@ export const __query = Object.assign(
     source = await loadDataSource(await source, "table");
     if (isDatabaseClient(source)) return evaluateQuery(source, makeQueryTemplate(operations, source), invalidation);
     if (isDataArray(source)) return __table(source, operations);
+    if (isArrowTable(source)) return __arrow(source, operations, invalidation);
```
mbostock, line 150: Rather than doing this here, I think we’ll want to have loadDataSource return the DuckDBClient for us, when the source is an Arrow file or an Arrow table. This way it’ll get cached; see getCachedDataSource in the other repo. Something like this:

```js
export async function loadDataSource(source, mode) {
  if (source instanceof FileAttachment) {
    if (mode === "table") {
      switch (source.mimeType) {
        case "text/csv": return source.csv({typed: true});
        case "text/tab-separated-values": return source.tsv({typed: true});
        case "application/json": return source.json();
      }
    }
    if (mode === "table" || mode === "sql") {
      switch (source.mimeType) {
        case "application/x-sqlite3": return source.sqlite();
      }
      if (/\.arrow$/i.test(source.name)) return DuckDBClient.of({__table: await source.arrow({version: 9})});
    }
    throw new Error(`unsupported file type: ${source.mimeType}`);
  }
  if ((mode === "table" || mode === "sql") && isArrowTable(source)) {
    return DuckDBClient.of({__table: source});
  }
  return source;
}
```

Also, this approach will make Arrow files work from SQL cells, too (using `__table` as the table name).

mbostock (changes requested):
Good questions! Here’s how to make it persist.

src/table.js
mbostock: (I do wonder whether we should try to create a better default name than `__table`; for example we could try to guess a table name from the name of the file attachment or the name of the cell, `source.name`. But then the data table cell will need to use the same logic to automatically choose the operations.from.table, so maybe it’s simpler just to use `__table`.)

libbey-observable: Thanks! That works, and yeah, makes sense as a better approach.

mbostock (approved):

src/table.js
libbey-observable: (And I do like the simplicity of `__table`)

PR #323, opened 2022-11-16
Export __table function

mkfreeman:
Exports the `__table` function for use in the Observable Data Table and SQL cells.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,11 @@
 export {FileAttachments, AbstractFile} from "./fileAttachment.js";
 export {Library} from "./library.js";
 export {getArrowTableSchema, isArrowTable} from "./arrow.js";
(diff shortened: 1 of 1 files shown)


mbostock (changes requested):
I think we shouldn’t export this under the name `__table`; that (internal) name was chosen for parity with `__query`, and the name `__query` was chosen because it needs to be exposed to notebooks (so that we can call it from the compiled representation of a SQL or data table cell).

But that’s not the case with `__table`. It’s never called directly from notebooks, so we don’t need to give it a dunderscore name.

Maybe a big clunky name like `applyDataTableOperations`?

mkfreeman:
Love a big clunky name (and thanks for the background context on the name)

mbostock (approved):

PR #324, opened 2022-11-16
use DuckDB for more data sources

mbostock:
This…

* Adopts DuckDB for CSV, TSV, JSON, Arrow and Parquet file sources for SQL cells
* Adopts DuckDB for Arrow and Parquet file sources for data table cells
* Adopts DuckDB for Arrow table sources for data table cells
* Fixes a bug with DuckDBClient.describeColumns when the table name needs escaping

🦆🦆🦆

### DIFF
--- a/src/arrow.js
+++ b/src/arrow.js
@@ -1,3 +1,6 @@
+import {arrow9 as arrow} from "./dependencies.js";
+import {cdn} from "./require.js";
+
 // Returns true if the vaue is an Apache Arrow table. This uses a “duck” test
 // (instead of strict instanceof) because we want it to work with a range of
 // Apache Arrow versions at least 7.0.0 or above.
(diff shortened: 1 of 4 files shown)


libbey-observable (approved):

mbostock:
Updated to pass the friendly name for data table cells, too. And also I dropped the table alias `t` from the generated SQL to make it prettier (especially since this one letter abbreviation isn’t likely to match the actual name of the table).

libbey-observable (approved):
All looks great, still testing locally for [9138](https://github.com/observablehq/observablehq/pull/9138), but think everything in here makes sense

mbostock:
Added support for primitive arrays in 1601b28. I also noticed that the DuckDBClient isn’t handling the FLOAT type correctly, so I think my assumption that DESCRIBE would always return the canonical type names (rather than aliases) listed here may not be true: https://duckdb.org/docs/sql/data_types/overview ☹️

mbostock:
I think we need to add caching here… otherwise we are re-creating the DuckDBClient every time we run a query.

libbey-observable:
Looked this over again, looks great!

PR #325, opened 2022-11-21
use DuckDB’s auto delimiter by default

mbostock:
Rather than hard-coding CSV = comma and TSV = tab (despite the name), use DuckDB’s magical automatic delimiter inference by default. You can still specify a delimiter explicitly if you use the _delimiter_ option.

Ref. https://observablehq.com/d/f11bdd2a9af411c5

### DIFF
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -173,7 +173,6 @@ async function insertFile(database, name, file, options) {
         return await connection.insertCSVFromPath(file.name, {
           name,
           schema: "main",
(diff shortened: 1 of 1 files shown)


Fil (approved):
This will be particularly useful for French users of Excel, since by default Excel localizes _csv_ with a semicolon delimiter (because it also uses commas instead of dots for decimal numbers…). The user can Select File > Options > Advanced to change that, but in any case we have awful lots of semicolons-separated-values files.

PR #328, opened 2022-11-29
Fix sql table only processes first yield

mootari:
When loadSource instantiates a DuckDB instance from an array data source, it caches the corresponding DuckDBClient instance using the array instance as key. When the data source yields subsequent updates, these are no longer processed, breaking the data flow between the data source and SQL cell.

This PR implements a stop-gap solution based on a suggestion by @Fil:
- we check for a .done property on the array (which we have on e.g. DuckDB data sources)
- it resolves to instantiate the client only when `.done !== false`

**Note that this does not solve the following case.** Given the data source
```js
data = {
  const result = [];
  yield (result.push({v: 1}), result);
  yield (result.push({v: 2}), result);
}
```
the following SQL cell will only list the first row:
```sql
select * from data
```

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -178,13 +178,19 @@ function sourceCache(loadSource) {
     if (!source) throw new Error("data source not found");
     let promise = cache.get(source);
     if (!promise) {
+      let resolve;
       // Warning: do not await here! We need to populate the cache synchronously.
-      promise = loadSource(source, name);
+      promise = (new Promise(_resolve => {
+        resolve = _resolve;
+      })).then(source => loadSource(source, name));
+      cache.set(promise, resolve);
       cache.set(source, promise);
     }
+    if (source.done !== false) cache.get(promise)(source);
     return promise;
   };
 }
+
 
 const loadTableDataSource = sourceCache(async (source, name) => {
   if (source instanceof FileAttachment) {


mbostock (changes requested):
Hmm. This presumes that someone will call loadDataSource again after the streaming data is done. If no one calls it, then the promise will never resolve; sourceCache has no way to “listen” to the streaming data itself and be notified when it is done.

Also, even when loadDataSource is called again and the data is finally done, in practice we don’t care about those earlier promises resolving since the downstream work has already been invalidated by the new streaming data.

So I think we want a different solution here, but I’m not sure what it is. Perhaps it’s a matter of bypassing the cache entirely for array sources that are not done?

```js
    if (!source) throw new Error("data source not found");
    if (source.done === false) return loadSource(source, name); // bypass cache
    let promise = cache.get(source);
```

mootari:
> Hmm. This presumes that someone will call loadDataSource again after the streaming data is done. If no one calls it, then the promise will never resolve; sourceCache has no way to “listen” to the streaming data itself and be notified when it is done.

We only delay resolving if .done exists and is explicitely set to false. Resolving can already happen on the first call (unless I messed up somewhere).

> Also, even when loadDataSource is called again and the data is finally done, in practice we don’t care about those earlier promises resolving since the downstream work has already been invalidated by the new streaming data.

We only ever deal with one promise though, and the idea is to avoid unnecessary computation at the cost of delaying the display of results. Otherwise we'd create new DuckDBClient instances for an increasingly larger array (at least that's how I understand it).

mootari:
Yes, if I understood @Fil correctly the intention was to fix the case for when another DatabaseClient conforming data source is used, in which case we'd expect there to be a last yield with .done set to true.

mbostock:
It would fix the problem with our current usage, but I don’t think it’s robust so I’d like to look for alternative solutions.

mootari:
@mbostock I feel I'm not familiar enough with the internal data table APIs to suggest a solution. Will leave this problem to @libbey-observable and you (and hopefully learn a lot in the process 🙏🏻 ).

Fil:
This discussion has merit, beyond the urgent need to fix. And the pairing session was enlightening!

PR #329, opened 2022-11-29
Update caching logic for tables

libbey-observable:
Resolves https://github.com/observablehq/observablehq/issues/9381

I did see the other possible solution suggestion, to bypass the cache entirely for array sources that are not done. If that seems like a better solution, happy to revise.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -177,9 +177,10 @@ function sourceCache(loadSource) {
   return (source, name) => {
     if (!source) throw new Error("data source not found");
     let promise = cache.get(source);
-    if (!promise) {
+    if (!promise || (isDataArray(source) && source.length !== promise._numRows)) {
       // Warning: do not await here! We need to populate the cache synchronously.
       promise = loadSource(source, name);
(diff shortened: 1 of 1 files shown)


Fil (approved):
Thanks! Seems a bit like what I tried, with the (welcome) additional check on the array type.

One consequence is that it might instantiate (then forget) duckdb databases quite a few times when large datasets stream in, ending up by wrecking the wasm (here I'm hand-waving a lot) with out of memory errors. I've seen this quite a bit when trying to debug.

We might think of a better solution in the future, but this is a simple enough change. Most importantly it fixes the outstanding bug, which for me seems like the top priority. I prefer an out-of-memory crash than returning wrong analysis.

So: let's merge, and continue the discussion and "real use" testing?

libbey-observable:
> Thanks! Seems a bit like what I tried, with the (welcome) additional check on the array type.

Yes, entirely your approach, with Mike B's comments incorporated. I experimented around a bit, but a better solution didn't immediately jump out at me. I'll go ahead and merge, especially since it's not automatically deployed, so there will still be a chance to revise before it goes out.

PR #331, opened 2022-11-30
Add missing filter options

mkfreeman:
This is a small code change that warrants a detailed explanation. For context, the `__table` function creates JavaScript array filters that mimic SQL filter functions. This PR is concerned with the way we think about NULL values. The SQL code that gets constructed by the Table cell reads:

```js
// If an "IS NULL" filter is applied to a column, add a SQL "IS NULL" statement
case "n":
        appendSql(` IS NULL`, args);
        return;
```

The corresponding JavaScript code currently reads:

```js
// If an "IS NULL" filter is applied to a column, filter down to JavaScript `null` OR undefined values
case "n": {
        source = source.filter((d) => d[column] == null)
```

While this makes sense at face value, **the concept of `null` means something different in JavaScript and SQL**. In SQL, _"A field with a NULL value is a field with no value._" ([source](https://www.w3schools.com/sql/sql_null_values.asp)). In JavaScript, `null` is a [primitive value](https://developer.mozilla.org/en-US/docs/Glossary/Null) which may appear in the dataset. 

Currently, the way the `__table` function checks for `null` values doesn't align with the way the Observable Data Table cell checks for `NULL/EMPTY`. When computing summaries for the Summary Charts, the Data Table cell uses the following check:

```js
 isMissing = (value) => {
  // Note that we use Number.isNaN because isNaN can return unexpected results
  // for non-numeric values.
  return value == null || value === "" || Number.isNaN(value);
};
```
Note, it checks if the value is `null` or `undefined` in the same way the standard library currently does, and also includes two additional checks: empty string (`value === ""`) and NaN `Number.isNaN(value)`. These checks importantly identify common JavaScript data patterns that are uncommon (or not possible?) in SQL.

A remaining question is: _should we make any additional changes to the SQL code to ensure comparable behaviors_? 

See [this notebook](https://observablehq.com/d/085731c6e8717bb7) for additional context and interactive examples.


mbostock:
We shouldn’t change the behavior of the existing `n` and `nn` operators. They have a clear meaning—is null and is not null—and the empty string is not the same as null in both SQL and JavaScript. Moreover, we shouldn’t change the definition of existing operators since doing so would retroactively affect existing notebooks (and it’s difficult to argue that this is merely fixing a bug).

I think if you want a query that matches NULL (null), the empty string, and NaN in JavaScript (and maybe also Invalid Date objects? I don’t recall how we represent those), it should probably be an `in` query, or perhaps an entirely new operator so that we can avoid changing the behavior of existing ones. I think it’s reasonable to map the JavaScript NaN value to be NULL in SQL, but I don’t think we should do the reverse because these are distinct types in JavaScript unlike SQL.

mkfreeman:
> _What are the corresponding SQL queries (if any)?_

Perhaps a missing `m` check for SQL is just `NULL` or `''` (empty string):

```js
case "m":
        appendSql(` IS NULL`, args);
        appendSql(` = ""`, args); // not sure if we need to use appendOperand before this
        break;
case "nm":
        appendSql(` IS NOT NULL`, args);
        appendSql(` !=""`, args);
```

mbostock:
The SQL for the `m` query would presumably have to look like `(x IS NULL OR x = '' OR x = float 'NaN')`?

(I’m tempted to use `d` for “defined” and `nd` for “not defined” to avoid a double negative, but it doesn’t matter much since those names are only used internally.)

mkfreeman:
A brief explanation of [Only check that the first value is a primitive type](https://github.com/observablehq/stdlib/pull/331/commits/b701de1e3042fe8ac3ef338554127aaf586a1776):

Because we want to support columns that put unmatched values into a "missing" bin, we no longer want to check that the first `n` (20) values are of the same type. For example, we want to support the array of strings `["a", "b", 1]` that indicates _1 value not string_. Without the above change, we would get an `Invalid table data` error pushed up.

mbostock (commented):
You appear to be removing some of the intentional “guard rails” for the data table cell here; this will allow the data table cell to be used on primitive arrays of mixed types (for example, a mix of Dates and objects, or a mix of booleans and numbers). Why do we need to do this? Is this type of mixed data something we expect to see in practice, and that we want to show in a data table cell? I’m not sure why we’re removing the consistency requirements that we previously implemented and that seem like a good idea (from my perspective).

mkfreeman:
Important questions! I wanted to [write out my thoughts](https://observablehq.com/d/bb4d481ef3b2ec34) before chatting. Feel free to comment here / in the notebook, or just wait until our meeting. 

I'd say we're just moving the guard rails somewhere else 😅 .

src/table.js
```diff
@@ -492,6 +475,24 @@ export function __table(source, operations) {
...
+          (d) => typeof d[column] === colType;
+        source = source.filter(filter);
+        break;
+      }
+      // not valid (doesn't match the column type)
+      case "nv": {
+        const [colType] = values;
+        const filter = colType === "date" ?
+          (d) => !(d[column] instanceof Date) || Number.isNaN(+d[column]) :
+          (d) => typeof d[column] !== colType;
```
mbostock, line 492: Can we have a shared function for `v` so that we can call that here and negate it? There are a number of subtleties (especially given the other comment) so it’d be good to not duplicate the logic.

src/table.js
```diff
@@ -492,6 +475,24 @@ export function __table(source, operations) {
     const [{value: column}] = operands;
     const values = operands.slice(1).map(({value}) => value);
     switch (type) {
+      // valid (matches the column type)
+      case "v": {
+        const [colType] = values;
+        const filter = colType === "date" ?
+          (d) => (d[column] instanceof Date) && !Number.isNaN(+d[column]) :
+          (d) => typeof d[column] === colType;
```
mbostock, line 483: TODO

* For the `date` type, stylistic simplification: `!isNaN(d[column])`
* For the `number` type, check isNaN too
* For the `buffer` type, this should check `d[column] instanceof ArrayBuffer`?
* For the `array` type, this should check `Array.isArray(d[column])`?
* For the `object` type, we’ll need to exclude null (because `typeof null` is `"object"`)
* For the `other` type, this should just check not-nullish?

mkfreeman:
Here's a [notebook](https://observablehq.com/d/c47a37c8e863bb3c) with the `isValid` function tested against a number of values. Two observations that I found initially surprising (but I think are fine): 
- `object` columns: `array`, `new ArrayBuffer`, and `date` are all valid (which makes sense, they're objects)
- `other` columns: everything except `null` and `undefined` is valid (makes sense! I guess they're valid forms of "other")

src/table.js
mkfreeman: Done! Added `isValid`

src/table.js
mkfreeman: Done in the last commit ✅ !

src/table.js
```diff
@@ -479,6 +462,27 @@ function likeOperand(operand) {
...
 
+// Check if a value is a valid form of the column type
+export function isValid(value, colType) {
+  switch (colType) {
+    case "string":
+    case "bigint":
+    case "boolean":
+      return typeof value === colType;
+    case "number":
+      return typeof value === colType && !Number.isNaN(value);
```
mbostock, line 473: You can simplify here, too, since at this point we already know that `value` is a number and hence will not require coerce.

```suggestion
      return typeof value === colType && !isNaN(value);
```

mbostock (commented):
Let me know when you’re ready for me to take a look.

src/table.js
mkfreeman: Maybe I'm mistaken, but I thought that `isNaN` does the coercion, so here is used `Number.isNaN` to _avoid_ unnecessarily doing coercion.

mbostock: That’s correct. You can use either here and get the same result. The only difference is that one is shorter to type—that’s all I was saying. You’re welcome to leave it as-is.

mkfreeman: Ah, by "simplify", I thought you meant simplify the computation. 

- `Number.isNaN` seems "simpler" in terms of the operation (doesn't do the additionally complex step of coercion)
- `isNaN` is simpler in terms of number of characters 😄

mkfreeman:
Great point -- appreciate you took the time to write that out, will make that change soon and re-request a review!

annie (approved):
LGTM based on the previous discussion in this PR!

perhaps we should also add some unit testing for `getValidator`, either in this or a later PR?

src/table.js
```diff
@@ -479,6 +479,42 @@ function likeOperand(operand) {
...
+const isValidBoolean = (value) => typeof value === "boolean";
+const isValidBigint = (value) => typeof value === "bigint";
+const isValidDate = (value) => value instanceof Date && !isNaN(value);
+const isValidBuffer = (value) => value instanceof ArrayBuffer;
+const isValidArray = (value) => Array.isArray(value);
+const isValidObject = (value) => typeof value === "object" && value !== null;
+const isValidOther = (value) => value != null;
+
+// Function to get the correct validity checking function based on type
+export function getValidator(colType) {
```
mbostock, line 494: I’d prefer a slightly more specific name for this if we’re going to expose it publicly. Maybe `getTypeValidator` would be sufficient?

mbostock (approved):

src/table.js
mkfreeman: I like `getTypeValidator`, and will add some tests!

test/table-test.js
```diff
@@ -542,3 +542,88 @@ describe("__table", () => {
...
+      {label: "array", value: [1, 2, 3]},
+      {label: "number", value: 10},
+      {label: "date", value: new Date(1)},
+      // eslint-disable-next-line no-undef
+      {label: "bigint", value: BigInt(10)},
+      {label: "null", value: null},
+      {label: "NaN", value: NaN},
+      {label: "undefined"}
+    ];
+  });
```
annie, line 563: i don't think you need to redefine `source` before every test case, since `.filter` creates a copy of the array instead of filtering in place.

```suggestion
  let source = [
    {label: "string", value: "string"},
    {label: "object", value: {}},
    {label: "buffer", value: new ArrayBuffer()},
    {label: "boolean", value: true},
    {label: "array", value: [1, 2, 3]},
    {label: "number", value: 10},
    {label: "date", value: new Date(1)},
    // eslint-disable-next-line no-undef
    {label: "bigint", value: BigInt(10)},
    {label: "null", value: null},
    {label: "NaN", value: NaN},
    {label: "undefined"}
   ];
```

annie: oh also why do you need the `// eslint-disable-next-line no-undef`?

mkfreeman: Thanks! Because this version of eslint doesn't recognize BigInt 🤷

PR #332, opened 2022-11-30
Support Arquero tables

### LINKED ISSUE #330: Arquero table support (stdlib)
Product issue: https://github.com/observablehq/observablehq/issues/9190

Update data source loading to support Arquero tables

libbey-observable:
Resolves #330 

- Identifies Arquero tables by presence of a `toArrow` function
- Uses Arrow's `.tableFromIPC()` with Arquero's `.toArrowBuffer()` as source for a new DuckDBClient

### DIFF
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -219,6 +219,8 @@ async function insertArrowTable(database, name, table, options) {
 }
 
 async function insertArqueroTable(database, name, source) {
+  // TODO When we have stdlib versioning and can upgrade Arquero to version 5,
+  // we can then call source.toArrow() directly, with insertArrowTable()
   const arrow = await loadArrow();
   const table = arrow.tableFromIPC(source.toArrowBuffer());
   return await insertArrowTable(database, name, table);
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -135,9 +135,9 @@ export class DuckDBClient {
           await insertArrowTable(db, name, source);
         } else if (Array.isArray(source)) { // bare array of objects
           await insertArray(db, name, source);
-        } else if(isArqueroTable(source)) {
+        } else if (isArqueroTable(source)) {
           await insertArqueroTable(db, name, source);
-        }  else if ("data" in source) { // data + options
+        } else if ("data" in source) { // data + options
           const {data, ...options} = source;
           if (isArrowTable(data)) {
             await insertArrowTable(db, name, data, options);
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -2,6 +2,7 @@ import {getArrowTableSchema, isArrowTable, loadArrow} from "./arrow.js";
 import {duckdb} from "./dependencies.js";
 import {FileAttachment} from "./fileAttachment.js";
 import {cdn} from "./require.js";
+import {isArqueroTable} from "./table.js";
 
 // Adapted from https://observablehq.com/@cmudig/duckdb-client
 // Copyright 2021 CMU Data Interaction Group
@@ -134,7 +135,9 @@ export class DuckDBClient {
           await insertArrowTable(db, name, source);
         } else if (Array.isArray(source)) { // bare array of objects
           await insertArray(db, name, source);
-        } else if ("data" in source) { // data + options
+        } else if(isArqueroTable(source)) {
+          await insertArqueroTable(db, name, source);
+        }  else if ("data" in source) { // data + options
           const {data, ...options} = source;
           if (isArrowTable(data)) {
             await insertArrowTable(db, name, data, options);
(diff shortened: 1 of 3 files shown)


mbostock:
It would be nice to add this directly to DuckDBClient so that you can use Arquero tables with DuckDBClient.of in addition to data table/SQL cells.

src/duckdb.js
```diff
@@ -134,7 +135,9 @@ export class DuckDBClient {
           await insertArrowTable(db, name, source);
         } else if (Array.isArray(source)) { // bare array of objects
           await insertArray(db, name, source);
-        } else if ("data" in source) { // data + options
+        } else if(isArqueroTable(source)) {
```
mbostock, line 138: Prettier:

```suggestion
        } else if (isArqueroTable(source)) {
```

src/duckdb.js
```diff
@@ -134,7 +135,9 @@ export class DuckDBClient {
...
+          await insertArqueroTable(db, name, source);
+        }  else if ("data" in source) { // data + options
```
mbostock, line 140: Prettier:

```suggestion
        } else if ("data" in source) { // data + options
```

src/table.js
```diff
@@ -141,6 +141,11 @@ function isTypedArray(value) {
   );
 }
 
+export function isArqueroTable(value) {
+  // Arquero tables have a `toArrow` function
+  return typeof value.toArrow === "function";
```
mbostock, line 146: This will crash if value is null or undefined, rather than returning false as expected. We should add `value &&` here like we do for isArrowTable.

More importantly, though, our implementation doesn’t use _value_.toArrow—we use _value_.toArrowBuffer. And if we’re going to use _value_.toArrowBuffer, we should check that _value_.toArrowBuffer exists here rather than checking _value_.toArrow. Arguably, we could check both if we think there’s a risk of some other object implementing a _value_.toArrowBuffer method that does something different (_i.e._, that doesn’t return an Arrow IPC array buffer).

mbostock (changes requested):

src/table.js
libbey-observable: Thank you, @mbostock! I should have caught all of those. Updated in https://github.com/observablehq/stdlib/pull/332/commits/705af99e5e8ab96c70589cb333ff352ab775e471 and https://github.com/observablehq/stdlib/pull/332/commits/7a84daf6ceab653cf5d126ef25b30d003d1a74a3.

src/duckdb.js
```diff
@@ -215,6 +218,12 @@ async function insertArrowTable(database, name, table, options) {
   }
 }
 
+async function insertArqueroTable(database, name, source) {
+  const arrow = await loadArrow();
+  const table = arrow.tableFromIPC(source.toArrowBuffer());
+  return await insertArrowTable(database, name, table);
```
mbostock, line 224: I checked how Arquero is implemented, and _source_.toArrowBuffer calls _source_.toArrow internally.

https://github.com/uwdata/arquero/blob/1c8f438df033026535e1c4f67b4f891fa6f8f29e/src/table/column-table.js#L306-L308

https://github.com/uwdata/arquero/blob/1c8f438df033026535e1c4f67b4f891fa6f8f29e/src/format/to-arrow.js#L6-L8

So, this code is currently converting an Arquero Table to an Arrow Table to an Array Buffer then back again to an Arrow Table. You can skip two conversions by calling _source_.toArrow directly (and as we discussed earlier in this review, that means we should now check for the presence of _source_.toArrow instead of _source_.toArrowBuffer).

```suggestion
  return await insertArrowTable(database, name, source.toArrow());
```

mbostock (changes requested):

src/duckdb.js
libbey-observable: Thanks for taking a look, Mike! I started out trying to use _source_.toArrow rather than _source_.toArrowBuffer, but ran into errors that way (see screenshot), and the result of `requestDatabaseTables` for the cell when constructed that way is an empty array. 
Screen Shot 2022-12-01 at 7 20 13 PM

mbostock: What version of Arquero are you using?

Our DuckDBClient depends on Apache Arrow version 9 or later, which means you’ll need to use Arquero version 5 or later. The version of Arquero in the Observable Standard library is currently 4.8.8 which uses Apache Arrow ^3.0.0.

So, you’ll need to load a more recent version of Arquero for this to work, and more generally, we need to implement standard library versioning so that we can upgrade Arquero to a more recent version in new notebooks.

libbey-observable: I'm using Arquero from our stdlib; my opinion is that this should work with the Arquero in our stdlib. I added a comment in the code about upgrading Arquero when we have standard library versioning, and I suggest we update the `insertArqueroTable` function then. (We may also want the current implementation around for older notebooks.)

mbostock (approved):
Makes sense. 👍

I hope we can get to stdlib versioning soon!

PR #338, opened 2022-12-08
Update __table sorting to handle missing values

annie:
Resolves https://github.com/observablehq/observablehq/issues/9498

Cribbed `ascendingDefined` from [d3.sort](https://github.com/d3/d3-array/blob/main/src/sort.js#L37-L39) (it's not exported).

I'm a bit on the fence about whether or not missing values should be sorted first when the order is descending 🤔 but perhaps we can try it this way and see what people think?


src/table.js
```diff
@@ -565,9 +569,9 @@ export function __table(source, operations) {
     }
   }
   for (const {column, direction} of reverse(operations.sort)) {
-    const compare = direction === "desc" ? descending : ascending;
+    const compare = (a, b) => ascendingDefined(a[column], b[column]);
     if (source === input) source = source.slice(); // defensive copy
-    source.sort((a, b) => compare(a[column], b[column]));
+    source.sort(direction === "desc" ? (a, b) => compare(b, a) : compare);
```
mbostock, line 574: Tiny optimization to avoid an additional wrapper function in the descending case by taking the same approach for both:

```suggestion
    source.sort(direction === "desc" ? (a, b) => ascendingDefined(b[column], a[column]) : (a, b) => ascendingDefined(a[column], b[column]));
```

mbostock (approved):
I’m okay with this, but I’d prefer to put nulls at the end in both cases rather than to put nulls first in the descending case. I think if you want to look at nulls specifically in the data table cell, you can click on the corresponding bar in the summary chart to filter. But it’s not a strong preference and we can always change our minds in the future, so not a blocker.

annie:
oookay after some more testing, i started to agree with @mbostock that sorting nulls at the end felt more intuitive, so i updated the PR to do that. let's see what people think!

annie:
@mbostock would you mind taking another look at this? i changed the implementation slightly to sort nulls at the end.

mbostock (approved):
👍👍

PR #340, opened 2022-12-16
Fix v and nv filters in SQL

annie:
This fixes a bug where `v` and `nv` filters don't work for SQL data sources 😬 The issue is that, because we handle those filters in the binary operation conditional block, we append the primitive value (which contains the expected type) to the SQL query https://github.com/observablehq/stdlib/blob/main/src/table.js#L420. You end up with a query with broken syntax in the `WHERE` clause:

Screen Shot 2022-12-16 at 13 06 59

We should just consider `v` and `nv` to be unary operations (I think it was originally my suggestion to handle them in the binary op block so that's my bad!!). I also added a unit test for this case.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -380,13 +380,17 @@ function appendWhereEntry({type, operands}, args, escaper) {
   if (operands.length < 1) throw new Error("Invalid operand length");
 
   // Unary operations
-  if (operands.length === 1) {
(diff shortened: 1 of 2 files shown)


mkfreeman (approved):
LGTM, thanks for catching that and adding a test!

PR #344, opened 2023-01-19
Infer schema for relevant data sources

libbey-observable:
In this approach, we take a sample (currently the first 100 rows), and for each column, count how many times we encounter each possible data type. Then we take the most frequently encountered type as the column's type. We could also add some random sampling.

**Note** that in the screenshots, the data in the columns has not been coerced, this is an in-between point, where we've inferred types, but not yet applied them.

From a CSV file:
Screen Shot 2023-01-19 at 3 51 20 PM

From a JSON file: 
Screen Shot 2023-01-19 at 3 36 49 PM

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -191,8 +191,8 @@ function sourceCache(loadSource) {
 const loadTableDataSource = sourceCache(async (source, name) => {
   if (source instanceof FileAttachment) {
     switch (source.mimeType) {
-      case "text/csv": return source.csv({typed: true});
-      case "text/tab-separated-values": return source.tsv({typed: true});
+      case "text/csv": return source.csv();
+      case "text/tab-separated-values": return source.tsv();
       case "application/json": return source.json();
       case "application/x-sqlite3": return source.sqlite();
     }
--- a/src/table.js
+++ b/src/table.js
@@ -693,7 +693,10 @@ function initKey() {
 function inferSchema(source) {
   const schema = [];
   const sampleSize = 100;
-  const sample = source.slice(0, sampleSize);
+  let sample = source.slice(0, sampleSize);
+  if (arrayIsPrimitive(sample)) {
+    sample = sample.map(d => {return {value: d};});
+  }
   const typeCounts = {};
   sample.map((d) => {
     for (const key in d) {
--- a/src/table.js
+++ b/src/table.js
@@ -543,8 +543,9 @@ export function getTypeValidator(colType) {
 // DuckDBClient for data arrays, too, and then we wouldn’t need our own __table
 // function to do table operations on in-memory data?
 export function __table(source, operations) {
-  const input = source;
   let {schema, columns} = source;
+  if (!schema) source.schema = inferSchema(source);
+  const input = source;
   let primitive = arrayIsPrimitive(source);
   if (primitive) source = Array.from(source, (value) => ({value}));
   for (const {type, operands} of operations.filter) {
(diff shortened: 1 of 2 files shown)


src/table.js
```diff
@@ -666,3 +667,79 @@ export function __table(source, operations) {
...
+        else if (
+          value.match(
+            /^([-+]\d{2})?\d{4}(-\d{2}(-\d{2})?)?(T\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/
+          )
+        )
+          typeCounts[key]["date"]++;
+        else if (value.match(/(\d{1,2})\/(\d{1,2})\/(\d{2,4}) (\d{2}):(\d{2})/))
+          typeCounts[key]["date"]++;
+        else if (value.match(/(\d{4})-(\d{1,2})-(\d{1,2})/))
+          typeCounts[key]["date"]++;
```
libbey-observable, line 720: Copied the lonnng regex from d3's autoType, and added a few more common date formats, but maybe we don't want to be that flexible?

mkfreeman: Are the types you added appropriately created as dates using `new Date()`? If so, I think that's fine (but maybe arbitrary....?).

mkfreeman:
This is looking great! For reference, here's how we do the random sampling for getting string lengths - it includes the first 20 rows (because they are what the user sees -- perhaps not necessary here), and randomly samples 100 values (using a seed so the random values are always the same). https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/computeSummaries.js#L86

src/table.js
libbey-observable: Yes, they do get created as the expected dates with `new Date()`.

mkfreeman: In that case, I'd say keep it!

src/table.js
```diff
@@ -543,8 +543,9 @@ export function getTypeValidator(colType) {
 // DuckDBClient for data arrays, too, and then we wouldn’t need our own __table
 // function to do table operations on in-memory data?
 export function __table(source, operations) {
-  const input = source;
   let {schema, columns} = source;
+  if (!schema || !isValidSchema(schema)) source.schema = inferSchema(source);
```
mbostock, line 547: This is mutating the source.schema which will be visible externally and we should avoid because mutation of inputs is an unexpected side-effect of calling this function.

If necessary we can use a WeakMap to instead associate sources with valid schemas if we don’t want to re-infer them repeatedly. (It might also make sense to move this schema inference earlier, say in loadDataSource? Not sure though.)

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+    });
+  }
+  const typeCounts = {};
+  sample.map((d) => {
+    for (const key in d) {
+      if (!typeCounts[key]) typeCounts[key] = initKey();
+      // for json and sqlite, we already have some types, but for csv and tsv, all
+      // columns are strings here.
+      const type = typeof d[key];
+      const value = type === "string" ? d[key]?.trim() : d[key];
```
mbostock, line 709: If `type` here is `"string"` then it’s not possible for `d[key]` to be nullish and hence we don’t need the optional chaining operator.

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+  const schema = [];
+  const sampleSize = 100;
+  let sample = source.slice(0, sampleSize);
+  if (arrayIsPrimitive(sample)) {
+    sample = sample.map((d) => {
+      return {value: d};
+    });
+  }
+  const typeCounts = {};
+  sample.map((d) => {
```
mbostock, line 703: The return value of sample.map is being ignored here. We should prefer a for-of loop instead as the standard mechanism for iteration (and if not that, then sample.forEach, but I usually try to use for-of for consistency).

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+    for (const key in d) {
+      if (!typeCounts[key]) typeCounts[key] = initKey();
+      // for json and sqlite, we already have some types, but for csv and tsv, all
+      // columns are strings here.
+      const type = typeof d[key];
+      const value = type === "string" ? d[key]?.trim() : d[key];
+      if (value === null || value === undefined || value.length === 0)
+        typeCounts[key]["other"]++;
```
mbostock, line 711: Not sure it’s appropriate to consider null/undefined/empty to be “other”. I would consider a column with 80% null/undefined and 20% string to be type string. In other words instead of counting these empty/missing values as a type, we should ignore them and only count types of present values. Otherwise sparsely populated columns will be considered “other” which is likely undesirable.

(Perhaps this is an alternative to the special treatment of “other” below, but I think for values that are truly “other” then we wouldn’t want that special treatment; the special treatment is only needed because we are considering nullish/empty values as “other” here.)

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+      else if (type !== "string") {
+        if (Array.isArray(value)) typeCounts[key]["array"]++;
```
mbostock, line 713: Nit: The `["array"]` style here is unusual. The common style would be `.array`.

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+        else if (value instanceof Date) typeCounts[key]["date"]++;
+        else if (value instanceof ArrayBuffer) typeCounts[key]["buffer"]++;
+        else if (type in typeCounts[key]) typeCounts[key][type]++; // number, bigint, boolean, or object
+      } else {
+        if (value === "true" || value === "false") typeCounts[key]["boolean"]++;
+        else if (!isNaN(+value) && /^-?[0-9]+$/.test(value))
```
mbostock, line 719: Number coercion happens implicitly when you call isNaN, so you can say `isNaN(value)` instead of `isNaN(+value)`. Would it be possible to only test number coercion once rather than doing it again below?

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+          typeCounts[key]["integer"]++;
+        else if (!isNaN(+value)) typeCounts[key]["number"]++;
+        else if (
+          value.match(
+            /^([-+]\d{2})?\d{4}(-\d{2}(-\d{2})?)?(T\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/
+          )
+        )
+          typeCounts[key]["date"]++;
+        else if (value.match(/(\d{1,2})\/(\d{1,2})\/(\d{2,4}) (\d{2}):(\d{2})/))
```
mbostock, line 728: It would be nice to combine these into a single refer if possible.

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+        else if (value.match(/(\d{4})-(\d{1,2})-(\d{1,2})/))
+          typeCounts[key]["date"]++;
+        else typeCounts[key]["string"]++;
+      }
+    }
+  });
+  const columns = Object.keys(typeCounts);
+  for (const col of columns) {
+    // sort descending so most commonly encoutered type is first
+    const typesSorted = Object.keys(typeCounts[col]).sort(function (a, b) {
```
mbostock, line 739: Nit: If we remove the special treatment of “other” here, we could probably use d3.greatest to get the most common type rather than needing the more expensive sort. Though it’s probably not noticeable since the set of possible types is small.

src/table.js
```diff
@@ -666,3 +667,90 @@ export function __table(source, operations) {
...
+        else if (value.match(/(\d{1,2})\/(\d{1,2})\/(\d{2,4}) (\d{2}):(\d{2})/))
+          typeCounts[key]["date"]++;
+        else if (value.match(/(\d{4})-(\d{1,2})-(\d{1,2})/))
+          typeCounts[key]["date"]++;
+        else typeCounts[key]["string"]++;
+      }
+    }
+  });
+  const columns = Object.keys(typeCounts);
+  for (const col of columns) {
```
mbostock, line 737: Any reason not to use `for (const col in typeCounts)` here instead?

mbostock (commented):
Some code review.

Let’s find a time to chat through this in real-time. I think we only want to do this on conjunction with type coercion; otherwise we are advertising types that won’t match the values.

libbey-observable:
@mbostock Thanks again for the valuable feedback – the issues you mentioned have been addressed in https://github.com/observablehq/stdlib/pull/346. Closing this PR in favor of that.

PR #345, opened 2023-01-25
Handle column renaming in makeQueryTemplate and __table

annie:
Implement column renaming in stdlib. Specifically, we now set column name overrides based on the `names` property in `node.data.operations`, which is added in [this PR](https://github.com/observablehq/observablehq/pull/10009).

For SQL data sources, we do this by modifying the `SELECT` clause. For in-memory data sources, we create a copy of the source data and modify it to use the new names.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -663,7 +663,7 @@ export function __table(source, operations) {
       Object.fromEntries(operations.select.columns.map((c) => [c, d[c]]))
     );
   }
-  if (operations.names) {
+  if (!primitive && operations.names) {
     const overridesByName = new Map(operations.names.map((n) => [n.column, n]));
     if (schema) {
       schema = schema.map((s) => {
--- a/src/table.js
+++ b/src/table.js
@@ -307,7 +307,11 @@ export function makeQueryTemplate(operations, source) {
     throw new Error("missing from table");
   if (select.columns && select.columns.length === 0)
     throw new Error("at least one column must be selected");
-  const columns = select.columns ? select.columns.map(escaper).join(", ") : "*";
+  const names = new Map(operations.names?.map(({column, name}) => [column, name]));
+  const columns = select.columns ? select.columns.map((column) =>  {
+    const override = names.get(column);
+    return override ? `${escaper(column)} AS ${escaper(override)}` : escaper(column);
+  }).join(", ") : "*";
   const args = [
     [`SELECT ${columns} FROM ${formatTable(from.table, escaper)}`]
   ];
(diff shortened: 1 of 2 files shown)


annie:
hm looks like CI is failing because `String.replaceAll` (which i added in the unit test) isn't available in Node 14. i'll update it to just use `.replace`.

mbostock (commented):
What do you think about making this part of the existing operations.select? E.g.,

```js
operations.select = {
  columns: [
    {column: "foo", as: "foo_alias"},
    {column: "bar", as: "bar_alias"}
  ]
};
```

and then we’d treat the existing operations as if `column` and `as` were the same? I.e., this


```js
operations.select = {
  columns: [
    "foo",
    "bar"
  ]
};
```

would be the same as

```js
operations.select = {
  columns: [
    {column: "foo", as: "foo"},
    {column: "bar", as: "bar"}
  ]
};
```

annie:
oh interesting! in my first pass i had added `names` to `select` as a new property, but that ended up feeling unwieldy – your suggestion seems cleaner though. i'll take a shot at updating it in the monorepo PR and see if i run into any issues.

annie:
@mbostock i ran into one major issue while trying to implement your suggestion, which is that it doesn't allow us to preserve the renamed name when a user hides the column. we could potentially get around this by adding a `hidden` flag to each `columns` object and toggling that when we hide a column, instead of removing the object from the array, but that would require a fairly significant set of changes and i'm not sure it's worth doing.

given that, i think i'd prefer to stick with this `operations.names` approach, unless you have strong objections and/or other ideas!

mbostock:
@annie does that mean we’re still selecting and returning columns that are hidden when the query is run?

annie:
no we don't actually select hidden columns, we just keep the name information for that column in `operations.names`. that allows us to still show the renamed name in the columns dropdown (where hidden columns are listed in a deselected state), and the column will preserve its renamed name if it's reselected.

mbostock:
@annie Got it. Sorry, still waking up! ☕ 👍

mbostock:
No further suggestions from me. I’ll leave the reviewing to @libbey-observable and @mkfreeman! Thanks y’all. 🙏

libbey-observable (approved):
This all makes sense and looks good to me!

src/table.js
```diff
@@ -659,6 +663,27 @@ export function __table(source, operations) {
       Object.fromEntries(operations.select.columns.map((c) => [c, d[c]]))
     );
   }
+  if (!primitive && operations.names) {
```
libbey-observable, line 666: Ah, nice!

annie: the real fix for the original issue you found is actually in [this change in Next](https://github.com/observablehq/observablehq/pull/10009/commits/bb017ddd0d86a6e777a420281bd41a4f08feef7e)! but while i was investigating i realized we should also add a check here.

libbey-observable: aaah! thanks for explaining that.

PR #346, opened 2023-01-25
Infer schemas and coerce data for table cells

libbey-observable:
Resolves https://github.com/observablehq/observablehq/issues/9862 and https://github.com/observablehq/observablehq/issues/9673

In this inference approach, we take a sample (currently the first 100 rows), and for each column, count how many times we encounter each possible data type. Then, if > 90% of the sampled values conform to that type, we take the most frequently encountered type as the column's type. If not, we use type "other."

These changes include some support for upcoming [column type assertions](https://github.com/observablehq/observablehq/pull/10133), but for the purposes of this PR, I'm testing against main. 

https://user-images.githubusercontent.com/111310561/214719295-8326f325-879b-4b8a-9d9f-2397b28951de.mov

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -9,6 +9,5 @@ export {
   isDataArray,
   isDatabaseClient,
   __table as applyDataTableOperations,
-  getTypeValidator,
-  inferSchema
+  getTypeValidator
 } from "./table.js";
--- a/src/table.js
+++ b/src/table.js
@@ -607,11 +607,13 @@ export function coerceToType(value, type, options = {}) {
     }
     case "array":
       if (Array.isArray(value)) return value;
-      return value && Array.isArray(Array.from(value))
-        ? Array.from(value)
-        : defaultValue;
-    case "buffer":
+      return [value];
     case "object":
+      // this will return true for everything except null, undefined, strings,
+      // numbers, boolean, and symbols, so may yield unexpected results.
+      if (typeof value === "object") return value;
+      return {value: value};
+    case "buffer":
     case "other":
     default:
       return value || value === 0 ? value : defaultValue;
--- a/src/table.js
+++ b/src/table.js
@@ -626,7 +626,7 @@ export function __table(source, operations) {
     source = source.map(d => coerceRow(d, types, schema));
   } else if (inferredSchema) {
     // Coerce data according to new schema, unless that happened due to
-    // operations.type, above. 
+    // operations.type, above.
     source = source.map(d => coerceRow(d, types, schema));
   }
   for (const {type, operands} of operations.filter) {
@@ -779,9 +779,8 @@ function coerceRow(object, types, schema) {
   return coerced;
 }
 
-function initKey() {
+function createTypeCount() {
   return {
-    other: 0,
     boolean: 0,
     integer: 0,
     number: 0,
@@ -790,9 +789,26 @@ function initKey() {
     array: 0,
     object: 0,
     bigint: 0,
-    buffer: 0
+    buffer: 0,
+    defined: 0
   };
 }
+
+// Caution: the order below matters! 🌶️ The first one that passes the ≥90% test
+// should be the one that we chose, and therefore these types should be listed
+// from most specific to least specific.
+const types = [
+  "boolean",
+  "integer",
+  "number",
+  "date",
+  "bigint",
+  "array",
+  "object",
+  "buffer",
+  "string" // should probably always be last since it’s least specific
+  // Note: "other" is intentionally omitted; it is handed specially!
+];
 
 // We need to show *all* keys present in the array of Objects
 function getAllKeys(rows) {
@@ -818,48 +834,49 @@ export function inferSchema(source, columns = getAllKeys(source)) {
   const sampleSize = 100;
   let sample = source.slice(0, sampleSize);
   const typeCounts = {};
+  for (const col of columns) typeCounts[col] = createTypeCount();
+  // TODO invert order of these loops?
   for (const d of sample) {
     for (const col of columns) {
-      if (!typeCounts[col]) typeCounts[col] = initKey();
-      const type = typeof d[col];
       let value = d[col];
+      if (value == null) continue;
+      const colCount = typeCounts[col];
+      const type = typeof value;
       if (type !== "string") {
-        if (Array.isArray(value)) ++typeCounts[col].array;
-        else if (value instanceof Date) ++typeCounts[col].date;
-        else if (value instanceof ArrayBuffer) ++typeCounts[col].buffer;
+        ++colCount.defined;
+        if (Array.isArray(value)) ++colCount.array;
+        else if (value instanceof Date) ++colCount.date;
+        else if (value instanceof ArrayBuffer) ++colCount.buffer;
         else if (type === "number") {
-          ++typeCounts[col].number;
-          if (Number.isInteger(value)) ++typeCounts[col].integer;
+          ++colCount.number;
+          if (Number.isInteger(value)) ++colCount.integer;
         }
         // bigint, boolean, or object
-        else if (type in typeCounts[col]) ++typeCounts[col][type];
-        else if (value !== null && value !== undefined) ++typeCounts[col].other;
+        else if (type in colCount) ++colCount[type];
       } else {
         value = value.trim();
+        if (!value) continue;
+        ++colCount.defined;
+        ++colCount.string;
         if (/^(true|false)$/i.test(value)) {
-          ++typeCounts[col].string;
-          ++typeCounts[col].boolean;
+          ++colCount.boolean;
         } else if (value && !isNaN(value)) {
-          ++typeCounts[col].number;
-          if (Number.isInteger(+value)) ++typeCounts[col].integer;
-        } else if (/^\d+n$/.test(value)) ++typeCounts[col].bigint;
-        else if (DATE_TEST.test(value)) ++typeCounts[col].date;
-        else if (value) ++typeCounts[col].string;
+          ++colCount.number;
+          if (Number.isInteger(+value)) ++colCount.integer;
+        } else if (/^\d+n$/.test(value)) ++colCount.bigint;
+        else if (DATE_TEST.test(value)) ++colCount.date;
       }
     }
   }
   for (const col in typeCounts) {
-    let type = greatest(Object.keys(typeCounts[col]), d => typeCounts[col][d]);
-    const numDefined = sample.filter(
-      (d) =>
-        !(
-          d[col] == null ||
-          (typeof d[col] === "string" && d[col].trim() === "")
-        )
-    ).length;
-    // If over 90% of the sampled data counted as this type, use it. Otherwise,
-    // use "other."
-    type = typeCounts[col][type] / numDefined >= 0.9 ? type : "other";
+    const colCount = typeCounts[col];
+    // Chose the type with the greatest count that is also ≥90%; or, if no type
+    // meets that criterion, fallback to other.
+    const minCount = colCount.defined * 0.9;
+    const type =
+      greatest(types, (type) =>
+        colCount[type] >= minCount ? colCount[type] : NaN
+      ) ?? "other";
     schema.push({
       name: col,
       type: type,
--- a/src/table.js
+++ b/src/table.js
@@ -798,7 +798,11 @@ export function inferSchema(source) {
         if (Array.isArray(value)) typeCounts[key].array++;
         else if (value instanceof Date) typeCounts[key].date++;
         else if (value instanceof ArrayBuffer) typeCounts[key].buffer++;
-        // number, bigint, boolean, or object
+        else if (type === "number") {
+          if (/^-?[0-9]+$/.test(value)) typeCounts[key].integer++;
+          else typeCounts[key].number++;
+        }
+        // bigint, boolean, or object
         else if (type in typeCounts[key]) typeCounts[key][type]++;
         else if (value !== null && value !== undefined) typeCounts[key].other++;
       } else {
--- a/src/table.js
+++ b/src/table.js
@@ -864,9 +864,6 @@ export function inferSchema(source, columns = getAllKeys(source)) {
         else if (DATE_TEST.test(value)) ++colCount.date;
       }
     }
-  }
-  for (const col in typeCounts) {
-    const colCount = typeCounts[col];
     // Chose the non-string, non-other type with the greatest count that is also
     // ≥90%; or if no such type meets that criterion, fallback to string if
     // ≥90%; and lastly fallback to other.
--- a/src/table.js
+++ b/src/table.js
@@ -620,7 +620,8 @@ export function __table(source, operations) {
   let {schema, columns} = source;
   let inferredSchema = false;
   if (!isQueryResultSetSchema(schema)) {
-    schema = inferSchema(source, columns);
+    if (arrayIsPrimitive(source)) schema = inferFromPrimitive(source, columns);
+    else schema = inferSchema(source, columns);
     inferredSchema = true;
   }
   let primitive = arrayIsPrimitive(source);
@@ -810,16 +811,18 @@ function getAllKeys(rows) {
   return Array.from(keys);
 }
 
+export function inferFromPrimitive(source) {
+  const primitiveSource = source.map((d) => {
+    return {value: d};
+  });
+ return inferSchema(primitiveSource, ["value"]);
+}
+
+
 export function inferSchema(source, columns = getAllKeys(source)) {
   const schema = [];
   const sampleSize = 100;
   let sample = source.slice(0, sampleSize);
-  if (arrayIsPrimitive(sample)) {
-    sample = sample.map((d) => {
-      return {value: d};
-    });
-    columns.push("value");
-  }
   const typeCounts = {};
   for (const d of sample) {
     for (const col of columns) {
--- a/src/table.js
+++ b/src/table.js
@@ -573,21 +573,21 @@ export function coerceToType(value, type) {
     case "bigint":
       return typeof value === "bigint" || value == null
         ? value
-        : Number.isInteger(typeof value === "string" && !value ? NaN : +value)
+        : Number.isInteger(typeof value === "string" && !value.trim() ? NaN : +value)
         ? BigInt(value) // eslint-disable-line no-undef
         : undefined;
     case "integer": // not a target type for coercion, but can be inferred
     case "number": {
       return typeof value === "number"
         ? value
-        : value == null || (typeof value === "string" && !value)
+        : value == null || (typeof value === "string" && !value.trim())
         ? NaN
         : Number(value);
     }
     case "date": {
       if (value instanceof Date || value == null) return value;
       if (typeof value === "number") return new Date(value);
-      if (typeof value === "string" && !value) return null;
+      if (typeof value === "string" && !value.trim()) return null;
       const trimValue = String(value).trim();
       return new Date(DATE_TEST.test(trimValue) ? trimValue : NaN);
     }
@@ -860,8 +860,7 @@ export function inferSchema(source, columns = getAllKeys(source)) {
         } else if (value && !isNaN(value)) {
           ++colCount.number;
           if (Number.isInteger(+value)) ++colCount.integer;
-        } else if (/^\d+n$/.test(value)) ++colCount.bigint;
-        else if (DATE_TEST.test(value)) ++colCount.date;
+        } else if (DATE_TEST.test(value)) ++colCount.date;
       }
     }
     // Chose the non-string, non-other type with the greatest count that is also
--- a/src/table.js
+++ b/src/table.js
@@ -582,7 +582,7 @@ export function coerceToType(value, type) {
     case "date": {
       if (value instanceof Date) return value;
       const trimValue = typeof value === "string" ? value.trim() : value;
-      return value && /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))?([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(trimValue)
+      return /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(trimValue)
         ? new Date(trimValue)
         : new Date("");
     }
@@ -857,8 +857,7 @@ function inferType(type, value) {
       else return "number";
     } else if (/^\d+n$/.test(value)) return "bigint";
     else if (
-      value &&
-      /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))?([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(value)
+      /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(value)
     )
       return "date";
     // the long regex accepts dates in the form of ISOString and
--- a/src/table.js
+++ b/src/table.js
@@ -819,9 +819,7 @@ export function inferSchema(source, columns = getAllKeys(source)) {
   for (const d of sample) {
     for (const col of columns) {
       if (!typeCounts[col]) typeCounts[col] = initKey();
-      const type = typeof d[col];
-      const value = type === "string" ? d[col].trim() : d[col];
-      typeCounts[col][inferType(type, value)]++;
+      typeCounts[col][inferType(d[col])]++;
     }
   }
   for (const col in typeCounts) {
@@ -839,7 +837,9 @@ export function inferSchema(source, columns = getAllKeys(source)) {
   return schema;
 }
 
-function inferType(type, value) {
+function inferType(colValue) {
+  const type = typeof colValue;
+  const value = type === "string" ? colValue.trim() : colValue;
   // for json and sqlite, we already have some types, but for csv and tsv, all
   // columns are strings here.
   const typedNonStrings = ["bigint", "boolean", "object"];
--- a/src/table.js
+++ b/src/table.js
@@ -564,7 +564,7 @@ export function coerceToType(value, type, options = {}) {
     case "bigint":
       return typeof value === "bigint"
         ? value
-        : !value || isNaN(value)
+        : !value || isNaN(value) || !Number.isInteger(+value)
         ? numberDefault
         // eslint-disable-next-line no-undef
         : BigInt(value);
@@ -808,7 +808,7 @@ export function inferSchema(source) {
         else if (value instanceof Date) typeCounts[key].date++;
         else if (value instanceof ArrayBuffer) typeCounts[key].buffer++;
         else if (type === "number") {
-          if (/^-?[0-9]+$/.test(value)) typeCounts[key].integer++;
+          if (Number.isInteger(+value)) typeCounts[key].integer++;
           else typeCounts[key].number++;
         }
         // bigint, boolean, or object
@@ -817,7 +817,7 @@ export function inferSchema(source) {
       } else {
         if (value === "true" || value === "false") typeCounts[key].boolean++;
         else if (value && !isNaN(value)) {
-          if (/^-?[0-9]+$/.test(value)) typeCounts[key].integer++;
+          if (Number.isInteger(+value)) typeCounts[key].integer++;
           else typeCounts[key].number++;
         } else if (
           value &&
--- a/src/table.js
+++ b/src/table.js
@@ -547,8 +547,8 @@ export function coerceToType(value, type, options = {}) {
   const defaultValue = options.soft ? value : null;
   const numberDefault = defaultValue === null ? NaN : defaultValue;
   switch (type) {
-    case "string":
-      return value === "string"
+    case "string": 
+      return typeof value === "string"
         ? value.trim()
         : value || value === 0
         ? value.toString()
--- a/src/table.js
+++ b/src/table.js
@@ -66,7 +66,12 @@ function objectHasEnumerableKeys(value) {
 }
 
 function isQueryResultSetSchema(schemas) {
-  return (Array.isArray(schemas) && schemas.every((s) => s && typeof s.name === "string" && typeof s.type === "string"));
+  return (
+    Array.isArray(schemas) &&
+    schemas.every(
+      (s) => s && typeof s.name === "string" && typeof s.type === "string"
+    )
+  );
 }
 
 function isQueryResultSetColumns(columns) {
--- a/src/table.js
+++ b/src/table.js
@@ -633,8 +633,9 @@ export function __table(source, operations) {
     for (const {name, type} of operations.type) {
       types.set(name, type);
       // update schema with user-selected type
+      if (schema === input.schema) schema = schema.slice(); // copy on write
       const colIndex = schema.findIndex((col) => col.name === name);
-      if (colIndex > -1) schema[colIndex] = {name, type};
+      if (colIndex > -1) schema[colIndex] = {...schema[colIndex], type};
     }
     source = source.map(d => coerceRow(d, types));
   }
--- a/src/table.js
+++ b/src/table.js
@@ -546,7 +546,6 @@ export function getTypeValidator(colType) {
 }
 
 export function coerceToType(value, type) {
-  const stringValue = typeof value === "string" ? value.trim() : value;
   switch (type) {
     case "string":
       return typeof value === "string" || value == null
@@ -586,14 +585,15 @@ export function coerceToType(value, type) {
     case "date": {
       if (value instanceof Date) return value;
       if (typeof value === "string") {
+        const trimValue = value.trim();
         let match;
         if (
-          (match = stringValue.match(
+          (match = trimValue.match(
             /^([-+]\d{2})?\d{4}(-\d{2}(-\d{2})?)?(T\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/
           ))
         ) {
           if (fixTz && !!match[4] && !match[7])
-            value = stringValue.replace(/-/g, "/").replace(/T/, " ");
+            value = trimValue.replace(/-/g, "/").replace(/T/, " ");
         }
       }
       return new Date(value);
--- a/src/table.js
+++ b/src/table.js
@@ -849,9 +849,16 @@ export function inferSchema(source) {
     }
   }
   for (const col in typeCounts) {
+    let type = greatest(Object.keys(typeCounts[col]), (d) => typeCounts[col][d]);
+    // If over 90% of the sampled data counted as this type, use it. Otherwise,
+    // use "other."
+    type =
+      typeCounts[col][type] / Math.min(source.length, sampleSize) >= 0.9
+        ? type
+        : "other";
     schema.push({
       name: col,
-      type: greatest(Object.keys(typeCounts[col]), (d) => typeCounts[col][d])
+      type: type
     });
   }
   return schema;
--- a/src/table.js
+++ b/src/table.js
@@ -850,7 +850,13 @@ export function inferSchema(source, columns = getAllKeys(source)) {
   }
   for (const col in typeCounts) {
     let type = greatest(Object.keys(typeCounts[col]), d => typeCounts[col][d]);
-    const numDefined = sample.filter(d => !(d[col] == null || d[col] === "")).length;
+    const numDefined = sample.filter(
+      (d) =>
+        !(
+          d[col] == null ||
+          (typeof d[col] === "string" && d[col].trim() === "")
+        )
+    ).length;
     // If over 90% of the sampled data counted as this type, use it. Otherwise,
     // use "other."
     type = typeCounts[col][type] / numDefined >= 0.9 ? type : "other";
--- a/src/table.js
+++ b/src/table.js
@@ -551,6 +551,9 @@ export function getTypeValidator(colType) {
   }
 }
 
+// Accepts dates in the form of ISOString and LocaleDateString, with or without time
+const DATE_TEST = /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/;
+
 export function coerceToType(value, type) {
   switch (type) {
     case "string":
@@ -582,7 +585,7 @@ export function coerceToType(value, type) {
     case "date": {
       if (value instanceof Date) return value;
       const trimValue = typeof value === "string" ? value.trim() : value;
-      return /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(trimValue)
+      return DATE_TEST.test(trimValue)
         ? new Date(trimValue)
         : new Date("");
     }
@@ -856,12 +859,7 @@ function inferType(type, value) {
       if (Number.isInteger(+value)) return "integer";
       else return "number";
     } else if (/^\d+n$/.test(value)) return "bigint";
-    else if (
-      /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(value)
-    )
-      return "date";
-    // the long regex accepts dates in the form of ISOString and
-    // LocaleDateString, with or without times
+    else if (DATE_TEST.test(value)) return "date";
     else if (value) return "string";
     else return "other";
   }
--- a/src/table.js
+++ b/src/table.js
@@ -826,16 +826,11 @@ export function inferSchema(source, columns = getAllKeys(source)) {
     }
   }
   for (const col in typeCounts) {
-    let type = greatest(
-      Object.keys(typeCounts[col]),
-      (d) => typeCounts[col][d]
-    );
+    let type = greatest(Object.keys(typeCounts[col]), d => typeCounts[col][d]);
+    const numDefined = sample.filter(d => !(d[col] == null || d[col] === "")).length;
     // If over 90% of the sampled data counted as this type, use it. Otherwise,
     // use "other."
-    type =
-      typeCounts[col][type] / Math.min(source.length, sampleSize) >= 0.9
-        ? type
-        : "other";
+    type = typeCounts[col][type] / numDefined >= 0.9 ? type : "other";
     schema.push({
       name: col,
       type: type,
--- a/src/table.js
+++ b/src/table.js
@@ -853,7 +853,7 @@ function inferType(type, value) {
       else return "number";
     }
     else if (typedNonStrings.includes(type)) return type;
-    else if (value !== null && value !== undefined) return "other";
+    else return "other";
   } else {
     if (value === "true" || value === "false") return "boolean";
     else if (value && !isNaN(value)) {
@@ -867,5 +867,6 @@ function inferType(type, value) {
     // the long regex accepts dates in the form of ISOString and
     // LocaleDateString, with or without times
     else if (value) return "string";
+    else return "other";
   }
 }
--- a/src/table.js
+++ b/src/table.js
@@ -840,10 +840,7 @@ export function inferSchema(source, columns = getAllKeys(source)) {
           else typeCounts[col].number++;
         } else if (/^\d+n$/.test(value)) typeCounts[col].bigint++;
         else if (
-          value &&
-          value.match(
-            /^(([-+]\d{2})?\d{4}(-\d{1,2}(-\d{1,2})?)|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))?([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/
-          )
+          /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2})?)|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))?([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(value)
         )
           typeCounts[col].date++;
         // the long regex accepts dates in the form of ISOString and
--- a/src/table.js
+++ b/src/table.js
@@ -591,7 +591,7 @@ export function coerceToType(value, type) {
     case "object":
     case "buffer":
     case "other":
-      return value || value === 0 ? value : null;
+      return value;
     default:
       throw new Error(`Unable to coerce to type: ${type}`);
   }
--- a/src/table.js
+++ b/src/table.js
@@ -829,7 +829,7 @@ export function inferSchema(source, columns = getAllKeys(source)) {
         else if (value instanceof ArrayBuffer) ++typeCounts[col].buffer;
         else if (type === "number") {
           ++typeCounts[col].number;
-          if (Number.isInteger(+value)) ++typeCounts[col].integer;
+          if (Number.isInteger(value)) ++typeCounts[col].integer;
         }
         // bigint, boolean, or object
         else if (type in typeCounts[col]) ++typeCounts[col][type];
--- a/src/table.js
+++ b/src/table.js
@@ -576,27 +576,16 @@ export function coerceToType(value, type) {
           BigInt(value);
     case "integer": // not a target type for coercion, but can be inferred
     case "number": {
-      return value === 0
+      return typeof value === "number"
         ? value
-        : !value || isNaN(value)
+        : value == null || value === ""
         ? NaN
         : Number(value);
     }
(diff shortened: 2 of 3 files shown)


src/table.js
```diff
@@ -538,15 +538,88 @@ export function getTypeValidator(colType) {
   }
 }
 
+export function coerceToType(value, type, options = {}) {
+  const defaultValue = options.soft ? value : null;
```
mbostock, line 542: In the case of numbers, we probably want the default value to be NaN (or undefined) instead of null. If you expect everything to be numbers by asserting that the type is number, and we sometimes provide nulls, then the nulls are likely to be silently coerced to zeroes. It’s usually better to coerce them to NaN so they are properly numbers, and/or undefined which implicitly gets coerced to NaN when you perform arithmetic, so that missing values are silently replaced with zeroes and potentially corrupting calculations.

For the same reason, we might want dates to be represented as undefined or new Date(NaN) (“invalid date”)…

libbey-observable: Thanks! Revised as shown below. 
Soft-coerce|Hard-coerce
-|-
Screen Shot 2023-01-26 at 10 33 00 AM|Screen Shot 2023-01-26 at 10 33 13 AM
Screen Shot 2023-01-26 at 11 07 19 AM|Screen Shot 2023-01-26 at 11 07 57 AM

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
...
+      return isNaN(date) ? dateDefault : date;
+    }
+    case "array":
+      if (Array.isArray(value)) return value;
+      return value && Array.isArray(Array.from(value))
+        ? Array.from(value)
+        : defaultValue;
+    case "buffer":
+    case "object":
+    case "other":
```
libbey-observable, line 602: Any suggestions for how else we could handle coercing to array, buffer, or other?

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
   }
 }
 
+export function coerceToType(value, type, options = {}) {
+  const defaultValue = options.soft ? value : null;
+  const numberDefault = defaultValue === null ? NaN : defaultValue;
+  switch (type) {
+    case "string":
+      return value === "string"
+        ? value.trim()
```
mkfreeman, line 552: Applying the `.trim()` operation on strings feels like a nice feature, but I'm just pausing to think if it's expected (it's already done in [d3.autoType](https://github.com/d3/d3-dsv/blob/6d96d6c0c70add67ec3869a22b3bbcd072ce328b/src/autoType.js#L3
), so not worth discussing at length).

As a user, is there ever a case when I might want to retain the whitespace in my strings (perhaps to identify rows with data entry errors...)? I probably would always want to remove them, but that feels like a data cleaning step that's being done silently. 

If we move forward with an option to do soft-coercion, perhaps that shouldn't trim....?

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
...
+        : value || value === 0
+        ? value.toString()
+        : defaultValue;
+    case "boolean":
+      return value === true || value === "true"
```
mkfreeman, line 557: Do we want to add `value.trim() === "true"`, so that the string `"true "` that includes whitespace is also coerced to true?

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
...
+        ? true
+        : value === false || value === "false"
```
mkfreeman, line 559: Same behavior here re supporting `.trim()`

src/table.js
mkfreeman: I'm sure you've considered this, but would it be expected to take the value and put it into an array or object?
```js
case "array":
   return [defaultValue]
case "object"
  return {value: defaultValue}
```

I imagine this could be useful when using the data table cell for data transformations (i.e., so that downstream values all have the specified type). Not yet super helpful for representing values in the table, as we don't have great displays for these types.

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
...
+    newlyInferred = true;
+  }
   let primitive = arrayIsPrimitive(source);
   if (primitive) source = Array.from(source, (value) => ({value}));
+  // Combine column types from schema with user-selected types in operations
+  const types = new Map(schema.map(({name, type}) => [name, type]));
+  if (operations.type) {
+    for (const {name, type} of operations.type) {
+      types.set(name, type);
+      source = source.map(d => coerceRow(d, types));
```
mkfreeman, line 627: Not sure if this is a performance concern, but here's we're iterating through the entire dataset for each user-specified type. Could we iterate through the all the types and perform `types.set(name, type);`, then afterward just iterate through `source = source.map(d => coerceRow(d, types));` one time?

mkfreeman:
In the (soon to be outdated) schema inference in `computeSummaries`, we iterate through all of the rows in the dataset to ensure that we get all the keys:

```js
// We need to show *all* keys present in the array of Objects
function getAllKeys(rows) {
  const keys = new Set();
  for (const row of rows) {
    // avoid crash if row is null or undefined
    if (row) {
      // only enumerable properties
      for (const key in row) {
        // only own properties
        if (Object.prototype.hasOwnProperty.call(row, key)) {
          // unique properties, in the order they appear
          keys.add(key);
        }
      }
    }
  }
  return Array.from(keys);
}

```

I don't see a comparable replacement for that in this PR - are we foregoing that step moving forward (and if so, how will we handle columns that don't get typed because they aren't present in the sample...)?

src/table.js
```diff
@@ -538,15 +543,99 @@ export function getTypeValidator(colType) {
...
+        ? true
+        : value === false || value === "false"
+        ? false
+        : defaultValue;
+    case "integer":
+      return !value || isNaN(parseInt(value)) ? numberDefault : parseInt(value);
+    case "bigint":
+      return typeof value === "bigint"
+        ? value
+        : !value || isNaN(value) || !Number.isInteger(+value)
```
mkfreeman, line 568: This `!value` check creates unexpected behavior for the number 0 (which I believe should be coercible to the BigInt `0n` via `BigInt(0)`. Is the `isNaN` check sufficient, or is `!value` catching certain values like `null` or `undefined`?

src/table.js
```diff
@@ -538,15 +543,100 @@ export function getTypeValidator(colType) {
...
+        ? numberDefault
+        // eslint-disable-next-line no-undef
+        : BigInt(value);
+    case "number": {
+      return !value || isNaN(value) ? numberDefault : Number(value);
```
mkfreeman, line 573: This `!value` also results in returning `NaN` for the number `0`. Seems like we should add a test that includes the number `0`.

src/table.js
libbey-observable: Thanks for all the valuable feedback, @mkfreeman! I've made revisions to fix almost everything you mentioned, and am going to think about the `getAllKeys` question. I'll also be making the revisions discussed in the team meeting, including removing soft coercion.

src/table.js
libbey-observable: We would want `value` rather than `defaultValue` (as `defaultValue` is only supposed to be returned in the event of coercion failure), but otherwise, sure, perhaps that's a fine way to handle them. Will update.

libbey-observable:
> In the (soon to be outdated) schema inference in `computeSummaries`, we iterate through all of the rows in the dataset to ensure that we get all the keys:
> I don't see a comparable replacement for that in this PR - are we foregoing that step moving forward 

Added, thanks @mkfreeman!

src/table.js
```diff
@@ -666,3 +762,118 @@ export function __table(source, operations) {
...
+          keys.add(key);
+        }
+      }
+    }
+  }
+  return Array.from(keys);
+}
+
+export function inferSchema(source) {
+  const allKeys = getAllKeys(source);
```
mbostock, line 816: If we already know which columns we are selecting (which we often, but not always, do) we could avoid this and look at operations.select instead.

src/table.js
```diff
@@ -538,15 +543,106 @@ export function getTypeValidator(colType) {
...
   let primitive = arrayIsPrimitive(source);
   if (primitive) source = Array.from(source, (value) => ({value}));
+  // Combine column types from schema with user-selected types in operations
+  const types = new Map(schema.map(({name, type}) => [name, type]));
+  if (operations.type) {
+    for (const {name, type} of operations.type) {
+      types.set(name, type);
+      // update schema with user-selected type
+      const colIndex = schema.findIndex((col) => col.name === name);
+      if (colIndex > -1) schema[colIndex] = {name, type};
```
mbostock, line 637: This is potentially mutating the schema that the source provides; we can only mutate this if we “own” the schema (i.e., we created it inside this function, as when we call inferSchema). Also, I think we want to preserve the `inferred` field here.

src/table.js
```diff
@@ -538,15 +543,106 @@ export function getTypeValidator(colType) {
...
+    }
+    source = source.map(d => coerceRow(d, types));
+  }
+  // Coerce data according to new schema, unless that happened due to
+  // operations.type, above. 
+  if (newlyInferred && !operations.type) {
```
mbostock, line 643:
```suggestion
  else if (newlyInferred) {
```

src/table.js
mbostock:
```js
if (schema === input.schema) schema = schema.slice(); // copy on write
```

mbostock: One way to preserve other fields:

```js
schema[colIndex] = {...schema[colIndex], type};
```

src/table.js
```diff
@@ -538,15 +543,106 @@ export function getTypeValidator(colType) {
...
+}
+
 // This function applies table cell operations to an in-memory table (array of
 // objects); it should be equivalent to the corresponding SQL query. TODO Use
 // DuckDBClient for data arrays, too, and then we wouldn’t need our own __table
 // function to do table operations on in-memory data?
 export function __table(source, operations) {
   const input = source;
   let {schema, columns} = source;
+  let newlyInferred = false;
```
mbostock, line 623: Maybe call this `inferredSchema`? Since we’ll re-infer this every time `__table` is called so there isn’t really a concept of “newly” here.

src/table.js
```diff
@@ -538,15 +543,106 @@ export function getTypeValidator(colType) {
...
+  if (!schema || !isQueryResultSetSchema(schema)) {
```
mbostock, line 624:
```suggestion
  if (!isQueryResultSetSchema(schema)) {
```

src/table.js
```diff
@@ -666,3 +762,118 @@ export function __table(source, operations) {
   }
   return source;
 }
+
+function coerceRow(object, types) {
+  let coerced = {};
```
mbostock, line 768:
```suggestion
  const coerced = {};
```

src/table.js
```diff
@@ -666,3 +762,118 @@ export function __table(source, operations) {
...
+  for (var key in object) {
```
mbostock, line 769: We shouldn’t use a for-in here; we should be using the schema to know which columns are defined (and in which order).

src/table.js
```diff
@@ -810,52 +810,51 @@ function getAllKeys(rows) {
...
+export function inferSchema(source, columns = getAllKeys(source)) {
   const schema = [];
   const sampleSize = 100;
   let sample = source.slice(0, sampleSize);
   if (arrayIsPrimitive(sample)) {
     sample = sample.map((d) => {
       return {value: d};
     });
-    allKeys.push("value");
+    columns.push("value");
```
mbostock, line 821: (This is now mutating an input. Perhaps we can move the promotion of primitive arrays to {value} objects to before we infer the schema, so that inferSchema doesn’t have to handle primitive arrays.)

src/index.js
```diff
@@ -9,5 +9,6 @@ export {
   isDataArray,
   isDatabaseClient,
   __table as applyDataTableOperations,
-  getTypeValidator
+  getTypeValidator,
+  inferSchema
```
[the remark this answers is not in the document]
libbey-observable: Discussed with Mike F, and removed this export from index.js.

mkfreeman:
We'll need to handle the case when a user asserts that an array of primitives is a type that, when coerced, returns all `null` values. For example, if the data is `d3.range(10)`, and the user selected the type `boolean`. 

This will currently break because `arrayIsPrimitive` will return `false`, and the worker won't convert the primitive values into objects:

```js
// This code path will not be entered in the case described above
let value = cellValue;
if (arrayIsPrimitive(cellValue)) { // not true if an array of numbers is set to boolean
  value = Array.from(cellValue, (d) => ({value: d}));
  value.schema = cellValue.schema;
}
```

I suggest we handle this at the stdlib level my adding the following code:

```js
export function arrayIsPrimitive(value) {
  return (
    isTypedArray(value) ||
    arrayContainsPrimitives(value) ||
    arrayContainsDates(value) ||
    arrayContainsNulls(value) // checks if all values are null
  );
}


// Given an array, checks that the first n elements are all null or undefined.
// This allows an array of all null or undefined values to be handled in the
// table cell, such as if a user asserts a type that coerces all array values to
// null
export function arrayContainsNulls(value) {
  const n = Math.min(nChecks, value.length);
  if (!(n > 0)) return false;
  return value.every(element => element == null)
}

```

src/table.js
```diff
@@ -66,7 +66,12 @@ function objectHasEnumerableKeys(value) {
 }
 
 function isQueryResultSetSchema(schemas) {
-  return (Array.isArray(schemas) && schemas.every((s) => s && typeof s.name === "string"));
+  return (
+    Array.isArray(schemas) &&
+    schemas.every(
+      (s) => s && typeof s.name === "string" && typeof s.type === "string"
```
mbostock, line 72: Totally optional, but for readability, we could break this out into a separate function:

```js
function isColumnSchema(schema) {
  return schema && typeof schema.name === "string" && typeof schema.type === "string"
}
```

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
   }
 }
 
+export function coerceToType(value, type) {
+  const stringValue = typeof value === "string" ? value.trim() : value;
```
mbostock, line 547: The name `stringValue` is a little confusing here because if `value` is not a string, then `stringValue` is not a string, too. It’s kind of like `maybeTrimValue`? But I think perhaps we can avoid this local variable entirely by changing some of the logic below, and thereby only trim if needed; see comments below.

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+  switch (type) {
+    case "string":
+      return typeof value === "string"
+        ? stringValue
+        : value || value === 0
+        ? value.toString()
+        : null;
```
mbostock, line 554: If I’m reading this correctly, this turns the boolean input value `false` into `null`; I expect the string `"false"` instead. It’s also surprising to me that we trim when coercing a string to a string; trimming seems appropriate when coercing to some _other_ type, e.g. to test for “true” and “false”, but I expect that if the input is already in the desired type the value should not be affected. Lastly I expect that we preserve undefined rather than coercing it to null.

In other words, what if we did this?

```suggestion
      return typeof value === "string" || value == null
        ? value
        : String(value);
```

Also, we should use `String(value)` instead of `value.toString()`; the latter can be overridden to return something that’s not a string, and we want to guarantee that the coerced value is a string (or null or undefined).

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+    case "boolean":
+      return value === true || stringValue === "true"
+        ? true
+        : value === false || stringValue === "false"
+        ? false
+        : null;
```
mbostock, line 560: In the case of the input _not_ being a string or a boolean, such as a number, I expect similar behavior to JavaScript’s boolean coercion. So rather than `0` and `1` both being coerced to `null`, I expect `0` to be coerced to `false` and `1` to be coerced to `true`. I likewise expect us to preserve undefined rather than coercing it to null.

```suggestion
      if (typeof value === "string") {
        const trimValue = value.trim();
        return trimValue === "true" ? true : trimValue === "false" ? false : null;
      }
      return typeof value === "boolean" || value == null
        ? value
        : Boolean(value);
```

I guess it’s a little inconsistent to use JavaScript’s boolean coercion for non-strings, and true/false testing for strings… but I think that’s unavoidable given that we’re using strings to represent “untyped” data.

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+    case "integer":
+      return value === 0
+        ? value
+        : !value || isNaN(parseInt(value))
+        ? NaN
+        : parseInt(value);
```
mbostock, line 566: I’d prefer to use `Number(value)` to coerce to a number rather than `parseInt` for two reasons. First, `parseInt` is looser: it ignores trailing non-whitespace characters, for example `parseInt(" 1 whatever ")` returns `1`, but `Number(" 1 whatever ")` returns `NaN`. Secondly, `parseInt` is going to be a lot slower if the input is _already_ numbers (or dates), since it’ll first coerce the value to a string and then parse it; it’s faster to coerce directly to a number.

I’m also not sure what our policy is on nullish values; should those be preserved or coerced to NaN? In JavaScript, `null` is coerced to `0`, but I could see us wanting to preserve nulls or coerce them to NaN. But in a case like the boolean `false` as input, I expect `0` rather than `NaN`.

I also wonder whether we need integer-specific coercion, or if `"integer"` should here be treated as an alias for `"number"` and do exactly the same thing. In other words, I don’t think it’s necessary for us to provide `"integer"` as a target type for coercion; it’s sufficient to provide `"number"`. (If you want to round your numbers to integers, you can do that yourself in JavaScript or SQL etc.)

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+        ) {
+          if (fixTz && !!match[4] && !match[7])
+            value = stringValue.replace(/-/g, "/").replace(/T/, " ");
+        }
+      }
+      return new Date(value);
+    }
+    case "array":
+      if (Array.isArray(value)) return value;
+      return [value];
```
mbostock, line 601: It feels surprising to me to wrap single values in an array, especially if that value is possibly nullish. I think we’d want to preserve null and undefined here, at least. Or perhaps we can avoid offering `"array"` as a type target for coercion? I don’t see that we need to offer it; it doesn’t make sense for CSV, which is our primary use case for coercion.

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+    case "object":
+      // this will return true for everything except null, undefined, strings,
+      // numbers, boolean, and symbols, so may yield unexpected results.
+      if (typeof value === "object") return value;
+      return {value: value};
```
mbostock, line 606: Same; wrapping as an object seems surprising to me. I think we shouldn‘t offer `"object"` as a target type for coercion.

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+    case "buffer":
+    case "other":
```
mbostock, line 608: Same here, I don’t think these should be offered as target types for coercion.

src/table.js
```diff
@@ -538,15 +543,105 @@ export function getTypeValidator(colType) {
...
+    default:
```
mbostock, line 609: We should probably throw an error if you attempt to coerce to a type that we don’t support.

mbostock (changes requested):
This is tricky! I have quite a few questions about the intended behavior of coercion. I am hoping we can pare down the scope here and have suggestions.

libbey-observable:
> We'll need to handle the case when a user asserts that an array of primitives is a type that, when coerced, returns all `null` values. For example, if the data is `d3.range(10)`, and the user selected the type `boolean`.
> 
> This will currently break because `arrayIsPrimitive` will return `false`, and the worker won't convert the primitive values into objects:

Thanks for raising this, @mkfreeman! I see what you mean about arrays of `null` values. If we want to treat those as arrays as primitives, we may want to stop ignoring them in the `arrayIsPrimitive` function, maybe something like what's shown below. But I'd be curious to hear @mbostock's thoughts on whether we do want to treat them that way? 

Screen Shot 2023-02-01 at 8 54 15 AM

mbostock:
I’ve been wondering if we should drop support for arrays of primitives since it seems to be adding a lot of complexity…

src/table.js
libbey-observable: I think if we offer `"integer"` as a target type for coercion, it would feel a bit broken if it acted as an alias for `"number"` and returned numbers with decimals. But I'm not opposed to removing it as a target type for coercion, and only offering `"number"`.

src/table.js
libbey-observable: Also agree that it would make sense to avoid offering `"array"`, `"object"`, `"buffer"`, and `"other"` as target types for coercion.

src/table.js
mbostock: Yeah, my preference is to remove `"integer"` as a coercion target type too.

src/table.js
mkfreeman: Agreed, I've updated the visible options for the user in this [draft PR](https://github.com/observablehq/observablehq/pu
Screen Shot 2023-02-01 at 1 05 53 PM
ll/10133).

libbey-observable: Thanks, @mkfreeman! Let's also remove "Integer" from that list.

mkfreeman:
> I’ve been wondering if we should drop support for arrays of primitives since it seems to be adding a lot of complexity…

As a DTC user, I feel strongly that we should continue supporting arrays of primitives, as it allows a quick visual summary of arrays (of primitives...).

In terms of your proposed solution @libbey-observable, I think we need `arrayContainsPrimitives` to return true if the values are a mix of a primitive type and null/undefined. We could support that via a different function (my suggestion above), or I think we'd need to tweak the `switch` statement to add a new  case for `undefined` (or the case when the type is `object` for `null` values), which feels a little tricky because we don't want to set the type to `null` unless they are all `null`.

src/table.js
mkfreeman: Happy to, but out of curiosity, is that because we don't expect users to need their data as integers, or because the coercion is tricky?

libbey-observable: I was agreeing with Mike B's comment above, that they can round values themselves in JavaScript or SQL, if that's what they want.

mkfreeman: Ah, missed that comment. Removing `interger` as a target!

annie:
(related to primitive arrays but sorry if this is too tangential for this PR) i was wondering why we coerce primitive arrays to `{value}` in the worker, if we're already doing it here? i see it in both [requestTableSummary](https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/table.js#L138-L140) and [requestTableRows](https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/table.js#L110-L112). is that code redundant, or does it cover some other case?

libbey-observable:
> i was wondering why we coerce primitive arrays to `{value}` in the worker, if we're already doing it here? i see it in both [requestTableSummary](https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/table.js#L138-L140) and [requestTableRows](https://github.com/observablehq/observablehq/blob/main/notebook-next/src/worker/table.js#L110-L112). is that code redundant, or does it cover some other case?

Thanks, @annie! I was actually just wondering the same thing. When I try removing those steps from worker/table.js, at least with the stdlib changes here (where an array of primitives gets promoted to objects on load), I'm not seeing any issues. I think we may be able to remove those, but would maybe wait until these stdlib changes are released, if we do go this route. And would test more.

src/table.js
libbey-observable: @mbostock: Thank you! I believe I've addressed the issues you raised. `"integer"`, `"array"`, `"object"`, `"buffer"`, and `"other"` still appear in the `switch` statement despite being removed as target types for coercion because they can still be inferred, so we need to handle them without throwing an error. 

I moved the promotion of primitive arrays to object arrays up into `loadTableDataSource`, and have not seen any issues, but if you have any thoughts on that, I'd love to hear.

src/table.js
```diff
@@ -538,15 +546,92 @@ export function getTypeValidator(colType) {
...
+        ? NaN
+        : // eslint-disable-next-line no-undef
+          BigInt(value);
+    case "integer": // not a target type for coercion, but can be inferred
+    case "number": {
+      return value === 0
+        ? value
+        : !value || isNaN(value)
+        ? NaN
+        : Number(value);
```
mbostock, line 579: I’m still a little uncertain on how we want to handle null/undefined, but how about this? In particular, this coerces `false` to `0` instead of `NaN`. Though, we might also want to prevent empty strings from getting coerced to `0` and instead coerce them to `NaN`?

```suggestion
      return typeof value === "number"
        ? value
        : value == null
        ? NaN
        : Number(value);
```

src/table.js
mbostock: I think for these unusual types we should just `return value` here and have coercion do nothing. I don’t think we want to convert `false` or the empty string `""` to `null`, for example.

src/table.js
```diff
@@ -659,10 +744,120 @@ export function __table(source, operations) {
...
+    const value = object[col.name];
+    coerced[col.name] = type === "raw" ? value : coerceToType(value, type);
+  }
+  return coerced;
+}
+
+// https://github.com/d3/d3-dsv/issues/45
+const fixTz =
+  new Date("2019-01-01T00:00").getHours() ||
+  new Date("2019-07-01T00:00").getHours();
```
mbostock, line 767: This bug has been fixed in Safari, so I think we could safely omit this workaround to simplify the code here.

src/table.js
libbey-observable: It does seem like we might want to coerce empty strings to `NaN` rather than `0`, so they won't throw off any computations.

src/table.js
libbey-observable: @mbostock: Thanks! Have made these updates.

src/table.js
```diff
@@ -542,15 +550,81 @@ export function getTypeValidator(colType) {
...
+      return typeof value === "number"
+        ? value
+        : value == null || value === ""
+        ? NaN
+        : Number(value);
+    }
+    case "date": {
+      if (value instanceof Date) return value;
+      const trimValue = typeof value === "string" ? value.trim() : value;
+      return new Date(trimValue);
```
mbostock, line 588: The behavior of the Date constructor is not consistent across browsers; some browsers (Chrome) are much more lenient than others (Safari). Maybe it’s okay to be as lenient as possible here, since we do a regex test when computing the inferred types for strings… but potentially we could repeat that regex test here, or perhaps use [isoformat’s parse](https://github.com/mbostock/isoformat#parsedate-fallback) (which we already use elsewhere) to only allow ISO 8601 formatted dates to be coerced.

src/table.js
```diff
@@ -684,10 +758,115 @@ export function __table(source, operations) {
...
+  const sampleSize = 100;
+  let sample = source.slice(0, sampleSize);
+  const typeCounts = {};
+  for (const d of sample) {
+    for (const col of columns) {
+      if (!typeCounts[col]) typeCounts[col] = initKey();
+      // for json and sqlite, we already have some types, but for csv and tsv, all
+      // columns are strings here.
+      const type = typeof d[col];
+      const value = type === "string" ? d[col].trim() : d[col];
```
mbostock, line 823: Purely a stylistic preference, but consider a form like this:

```js
// Given an arbitrary value, returns "array", "date", "number", "buffer", etc.
function inferType(value) {
```

Then here could say

```js
typeCounts[col][inferType(value)]++;
```

instead of needing separate statements for each type, e.g., instead of

```js
typeCounts[col].array++;
typeCounts[col].date++;
typeCounts[col].buffer++;
// etc.
```

Also, the `inferType` could be exported and tested directly from unit tests, which could make it easier to test.

(I further imagine that `inferType` could return undefined if given a nullish value, or `"null"` or `"undefined"`, or could you check whether the value is nullish before calling `inferType`…)

src/table.js
```diff
@@ -684,10 +758,115 @@ export function __table(source, operations) {
...
+        if (value === "true" || value === "false") typeCounts[col].boolean++;
+        else if (value && !isNaN(value)) {
+          if (Number.isInteger(+value)) typeCounts[col].integer++;
+          else typeCounts[col].number++;
+        } else if (/^\d+n$/.test(value)) typeCounts[col].bigint++;
+        else if (
+          value &&
+          value.match(
+            /^(([-+]\d{2})?\d{4}(-\d{1,2}(-\d{1,2})?)|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))?([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/
+          )
```
mbostock, line 845: We should use `/regex/.test(value)` here instead of `value.match(/regex/)`. It’s faster since it doesn’t need to return any information about matched groups. Also, you don’t need the `value &&` here since you’ve already checked that `typeof value === "string"`.

src/table.js
```diff
@@ -684,10 +758,115 @@ export function __table(source, operations) {
...
+    let type = greatest(
+      Object.keys(typeCounts[col]),
+      (d) => typeCounts[col][d]
+    );
+    // If over 90% of the sampled data counted as this type, use it. Otherwise,
+    // use "other."
+    type =
+      typeCounts[col][type] / Math.min(source.length, sampleSize) >= 0.9
+        ? type
+        : "other";
```
mbostock, line 864: This is requiring 90% of the sampled values to match the most common type, but we’re including null/undefined values in the denominator. I wonder if that’s being overly strict. For example, if we sample 100 values, 50 of them are dates and 50 of them are null, it feels reasonable to say that the inferred type is date?

src/table.js
```diff
@@ -542,15 +550,81 @@ export function getTypeValidator(colType) {
...
+  }
+  // Combine column types from schema with user-selected types in operations
+  const types = new Map(schema.map(({name, type}) => [name, type]));
+  if (operations.type) {
+    for (const {name, type} of operations.type) {
+      types.set(name, type);
+      // update schema with user-selected type
+      if (schema === input.schema) schema = schema.slice(); // copy on write
+      const colIndex = schema.findIndex((col) => col.name === name);
+      if (colIndex > -1) schema[colIndex] = {...schema[colIndex], type};
```
mbostock, line 620: [optional] As an alternative to generating `{name, type, inferred}` in `inferSchema`, we could preserve the original type here when applying type coercion:

```suggestion
      if (colIndex > -1) schema[colIndex] = {name, originalType: schema[colIndex].type, type};
```

That might be useful if we want to preserve the original type when applying type coercion even in the case where the schema is not inferred. I don’t recall if we ended up allowing type coercion/assertion in the case where a schema is already explicitly provided by the data source though. But since you’ve already done some downstream work using this schema feel free to ignore this suggestion if it’s not helpful!

src/table.js
```diff
@@ -542,15 +550,81 @@ export function getTypeValidator(colType) {
...
+    case "bigint":
+      return typeof value === "bigint"
+        ? value
+        : value === 0 || value === true || value === false
+        ? // eslint-disable-next-line no-undef
+          BigInt(value)
+        : !value || isNaN(value) || !Number.isInteger(+value)
+        ? NaN
+        : // eslint-disable-next-line no-undef
+          BigInt(value);
```
mbostock, line 576: I find the `value === 0` and `!value` tests difficult to follow given the myriad combinations of possible inputs. (For example, I’m thinking about empty strings, whitespace strings, objects that implement a valueOf method that returns zero or the empty string, etc.)

So, here is my attempt at rethinking it based on some of the ideas we discussed previously:

* Propagate bigints, null, and undefined as-is
* Otherwise, first coerce the input value to a number
* But, special-case falsey strings (empty/whitespace) to NaN rather than zero
* Then test if the number is an integer to avoid BigInt throwing an error
* Use undefined instead of NaN to represent missing data

```suggestion
      return typeof value === "bigint" || value == null
        ? value
        : Number.isInteger(typeof value === "string" && !value ? NaN : +value)
        ? BigInt(value) // eslint-disable-line no-undef
        : undefined;
```

The use of undefined is because NaN is a number type, not a BigInt type. So while it’s appropriate (even necessary to avoid a subsequent silent coercion of null to zero when using typed arrays) to use NaN to represent invalid data in the number case, I don’t think it applies here.

mbostock (changes requested):

src/table.js
libbey-observable: I like the idea of being as lenient as we can, so went with repeating the regex test here instead of using isoformat's parse (and I also revised the regex slightly after testing in Safari). But then, thinking that we want to return `Invalid Date` objects for anything that fails coercion, I added the fallback of returning new Date(""), which seems a little funky. (The `value &&` was needed because otherwise null gets coerced to Wed Dec 31 1969 16:00:00
GMT-0800 (Pacific Standard Time))

src/table.js
libbey-observable: Initially I had some resistance to this, but over the course of trying it out, was won over. Tomorrow I'll add some unit tests for `inferType`. (I think there's value in keeping the unit tests for `inferSchema` that are already in place.) I went with having `inferType` return `"other"` for undefined or nullish values, and I prefer that, but are you thinking of reasons not to do so?

src/table.js
libbey-observable: Yes! Updated to use only non-nullish and non-empty string values in the denominator.

src/table.js
libbey-observable: This is a good thought, but I think it makes sense to keep the association with `inferred`, at least for now, because the downstream use for it is restricted to sources we've inferred a schema for.

src/table.js
libbey-observable: Makes sense, thanks for thinking all that through and detailing it. Same for all this detailed feedback: much appreciated! 🙏

src/table.js
```diff
@@ -542,15 +551,79 @@ export function getTypeValidator(colType) {
...
+    case "integer": // not a target type for coercion, but can be inferred
+    case "number": {
+      return typeof value === "number"
+        ? value
+        : value == null || (typeof value === "string" && !value)
+        ? NaN
+        : Number(value);
+    }
+    case "date": {
+      if (value instanceof Date) return value;
```
mbostock, line 583: I’m strongly tempted to preserve null and undefined here, too.

```suggestion
      if (value instanceof Date || value == null) return value;
```

I understand why we replace null and undefined with NaN for the number type—that’s because we may want to use it in a typed array which doesn’t allow null (and undefined), and we don’t want to null to be coerced to zero because that loses information about missing data. But for date, it’s not clear to me that an invalid Date instance is “better” or “more expected” than null or undefined. What’s principle should we use?

I’m also tempted to add support for coercing numbers to dates; we already support the reverse of coercing dates to numbers implicitly, and it’s really easy:

```js
      if (typeof value === "number") return new Date(value);
```

src/table.js
```diff
@@ -542,15 +551,79 @@ export function getTypeValidator(colType) {
...
+      const trimValue = typeof value === "string" ? value.trim() : value;
+      return /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(trimValue)
+        ? new Date(trimValue)
+        : new Date("");
```
mbostock, line 587: Assuming we handle null, undefined, and numbers as above, I think at this point it makes sense to explicitly coerce the value to a string. We’re doing that _implicitly_ when we test against the regex, but I think it’d be better to do it _explicitly_ like so:

```suggestion
      const trimValue = String(value).trim();
      return new Date(
        /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(
          trimValue
        )
          ? trimValue
          : NaN
      );
```

Alternatively if we want to return undefined here instead of invalid dates:

```suggestion
      const trimValue = String(value).trim();
      return /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(
        trimValue
      )
        ? new Date(trimValue)
        : undefined;
```

src/table.js
```diff
@@ -684,10 +757,112 @@ export function __table(source, operations) {
...
+export function inferSchema(source, columns = getAllKeys(source)) {
+  const schema = [];
+  const sampleSize = 100;
+  let sample = source.slice(0, sampleSize);
+  const typeCounts = {};
+  for (const d of sample) {
+    for (const col of columns) {
+      if (!typeCounts[col]) typeCounts[col] = initKey();
+      const type = typeof d[col];
+      const value = type === "string" ? d[col].trim() : d[col];
```
mbostock, line 820: I think this trim should be moved to within inferType rather than here; it’s not needed in many cases, and I think it’ll be clearer when the trim is more closely associated with the logic that ignores whitespace (e.g., testing against `"true"` and `"false"`).

src/table.js
```diff
@@ -684,10 +757,112 @@ export function __table(source, operations) {
...
+    else if (typedNonStrings.includes(type)) return type;
+    else return "other";
+  } else {
+    if (value === "true" || value === "false") return "boolean";
+    else if (value && !isNaN(value)) {
+      if (Number.isInteger(+value)) return "integer";
+      else return "number";
+    } else if (/^\d+n$/.test(value)) return "bigint";
+    else if (
+      /^(([-+]\d{2})?\d{4}(-\d{2}(-\d{2}))|(\d{1,2})\/(\d{1,2})\/(\d{2,4}))([T ]\d{2}:\d{2}(:\d{2}(\.\d{3})?)?(Z|[-+]\d{2}:\d{2})?)?$/.test(value)
```
mbostock, line 860: Can we pull this regex out to a constant so that we don’t need to repeat it twice?

mbostock (commented):
Almost there! Just a few more things, and we’re done, I swear. 😅

src/table.js
mkfreeman: Chiming in briefly - (hope it's helpful!) - IMO, if a value isn't coercible to a date, we should return an invalid Date object to signal to users that a row isn't in the expected format. If a user saw `undefined`, they may think that the row is empty, rather than improperly formatted. 

Also, sorry if I missed this in an earlier conversation, but couldn't we just return `new Date(value)` in any case that `value` is defined? Then the `Date` constructor will handle returning `invalid Date`, and will also coerce numbers to dates (which feels appropriate). I briefly tested with:

```js
// Date constructor returns invalid date for incorrect formats:
new Date("what")
new Date([])
new Date({})
new Date(new ArrayBuffer)

// This is maybe expected, treating `true` as 1
new Date(true) // Wed Dec 31 1969 19:00:00 GMT-0500 (Eastern Standard Time)

// This I find surprising - that 0 and 1 return the same time 🤔 
new Date(false) // Wed Dec 31 1969 19:00:00 GMT-0500 (Eastern Standard Time)
```

src/table.js
```diff
@@ -542,15 +551,79 @@ export function getTypeValidator(colType) {
...
+export function coerceToType(value, type) {
+  switch (type) {
+    case "string":
+      return typeof value === "string" || value == null
+        ? value
+        : String(value);
+    case "boolean":
+      if (typeof value === "string") {
+        const trimValue = value.trim();
+        return trimValue === "true" ? true : trimValue === "false" ? false : null;
```
mkfreeman, line 566: Just want to note that I was surprised that our boolean coercion returned `false` for a column of string values, but `true` for other types, such as a column of empty objects. _Maybe_ the default return value for strings should be `Boolean(value)`, but then that ruins detecting the strings "true" and "false". 
Screen Shot 2023-02-02 at 11 45 14 AM

src/table.js
```diff
@@ -684,10 +760,107 @@ export function __table(source, operations) {
...
+    else if (value instanceof Date) return "date";
+    else if (value instanceof ArrayBuffer) return "buffer";
+    else if (type === "number") {
+      if (Number.isInteger(+value)) return "integer";
+      else return "number";
+    }
+    else if (typedNonStrings.includes(type)) return type;
+    else return "other";
+  } else {
+    if (value === "true" || value === "false") return "boolean";
```
mkfreeman, line 857: This edge case may not be worth considering, but if you have a mix of `"true"` and `"False"` (note the capitalization), you get an inferred type of "other". Not sure if it's worth also applying `.toLowerCase`?

Screen Shot 2023-02-02 at 11 52 41 AM

src/table.js
mbostock: > I went with having inferType return "other" for undefined or nullish values, and I prefer that, but are you thinking of reasons not to do so?

Yeah, I don’t think we want to return the `"other"` type for nullish values because it confounds types; `"other"` essentially means an unknown type (some weird type that’s not covered by the other types), and nullish values aren’t weird—they’re common. And pragmatically, to come back to the other comment https://github.com/observablehq/stdlib/pull/346#discussion_r1093733883, if we had 60% nullish values and 40% dates, I expect the inferred type of the column to be `"date"` rather than `"other"`. (I’m commenting without having read the latest version of the code.)

src/table.js
mbostock: The reason we don’t want to “`new Date(value)` in any case” is mentioned in https://github.com/observablehq/stdlib/pull/346/files/6bee09b37b3b0a32213203699a82bf306d7e17ae#diff-6e1c3dc6cd2f88ca61d65a2f261fd5adfb15f103d719af5acaa064433d3b7bee: 

> The behavior of the Date constructor is not consistent across browsers; some browsers (Chrome) are much more lenient than others (Safari). Maybe it’s okay to be as lenient as possible here, since we do a regex test when computing the inferred types for strings… but potentially we could repeat that regex test here, or perhaps use [isoformat’s parse](https://github.com/mbostock/isoformat#parsedate-fallback) (which we already use elsewhere) to only allow ISO 8601 formatted dates to be coerced.

I like the principle of distinguishing between missing and invalid/non-coercible values. That suggests preserving nullish values when coercing to dates, coercing the empty string to null, and coercing non-ISO 8601 date strings to “Invalid Date” objects.

src/table.js
mbostock: I expect the inferred type of `"string"` here (or `"boolean"` if we support case-insensitive coercion of `"False"` to `false`, but I don’t think that’s a requirement).

src/table.js
libbey-observable: For null and undefined, I agree, I don't see a benefit to returning an invalid Date instance. I think this is what you're saying below, too, @mkfreeman: that if a value is defined and can't be coerced, that's when an invalid Date instance makes sense/may be helpful. But if the value is nullish, I'm not able to think of a reason we would want to return an invalid Date instance for that. Let's use "the principle of least surprise"?? Also agree about adding support for coercing numbers to dates. Thanks for bringing this up, @mbostock.

And @mkfreeman, I initially had the same question: why not pass any defined value to `new Date()`, and let it handle returning valid or invalid Dates. But by using the same regex test we use for inference, we avoid discrepancies between what gets inferred as a date and what gets coerced to a date. (I have a local example of when this can be weird, and would be happy to demo.)

src/table.js
libbey-observable: I'm in favor of adding case-insensitive boolean inference/coercion, because of this example. The split between inferred booleans and strings means that neither breaks 90%, so we fall back to "other."

src/table.js
libbey-observable: That makes sense. I updated to only return "other" when `if (value)` passes.

src/table.js
mbostock: > The split between inferred booleans and strings means that neither breaks 90%, so we fall back to "other."

I still think `"string"` is more appropriate in this case because it is a superset; the input value `"true"` is both a valid boolean and a valid string.

libbey-observable: Definitely agree that `"string"` is more appropriate, but we'd quickly get into complexity if we tried adding more interpretation and handling on top of the type count results. In this case we ended up with `{boolean: 5, string: 5}`. (but adding `toLowerCase` does fix that, and seems like a sensible thing to support.)

src/table.js
libbey-observable: It returned `null` for string values (rather than `false`). I'm not sure what the right response is when a Date, object, or other types are coerced to boolean. `TRUE` seems surprising to me, but if that's what we want to do, then it does seem strange to have type `string` alone not conform to that. Any thoughts, @mbostock?

src/table.js
mkfreeman: Thanks for clarifying!

src/table.js
mbostock: I think we’ll have to solve this problem. There’s a similar issue with integer vs. number. For example, imagine you have the dataset [0.1, 0.2, 0.3, 0.4, 0.5, 1, 2, 3, 4, 5]: half are `"integer"` and half are `"number"`, but neither represents ≥90% and so the inferred type is `"other"`. In this case we should infer the type as `"number"` because an integer is a subtype of number. Here’s a (currently failing) test we can use:

```js
  it("infers mixed integers and numbers as numbers", () => {
    assert.deepStrictEqual(
      inferSchema([0.1, 0.2, 0.3, 0.4, 0.5, 1, 2, 3, 4, 5].map((x) => ({x}))),
      [{name: "x", type: "number", inferred: "number"}]
    );
  });
```

I guess what this means is that we can use the inferType(value) pattern (sorry!) and instead we have to support the idea that values can have multiple possible types. So the type counts for this column would look like `{integer: 5, number: 10}` and the totals of _all_ type counts could be greater than the number of (sampled) values, and we’d still use the greatest type count to infer the type (and return `"other"` if the greatest type count is <90%).

libbey-observable: Oooh, interesting! Gonna give this a try. 🙏

mbostock: Here’s a test for the true/false/other string case, too (using [PoltiFact’s Truth-o-Meter ratings](https://www.politifact.com/truth-o-meter/)):

```js
  it("infers boolean-ish strings and strings as strings", () => {
    assert.deepStrictEqual(
      inferSchema(["true", "false", "pants on fire"].map((x) => ({x}))),
      [{name: "x", type: "string", inferred: "string"}]
    );
  });
```

libbey-observable: This is a nice solution, thanks for that, @mbostock. Pushed this update, and seems to work as desired, although one interesting edge case is that `NaN` gets counted as a number, so if you have a mix of integers and `NaN`s, it will infer a type of `"number"` rather then `"integer"`.  Doesn't seem like a show-stopping issue, but thought I'd mention.

Screen Shot 2023-02-02 at 12 17 51 PM

src/table.js
```diff
@@ -542,15 +551,84 @@ export function getTypeValidator(colType) {
...
+export function coerceToType(value, type) {
+  switch (type) {
+    case "string":
+      return typeof value === "string" || value == null ? value : String(value);
+    case "boolean":
+      if (typeof value === "string") {
+        const trimValue = value.trim();
+        return trimValue.toLowerCase() === "true"
+          ? true
+          : trimValue.toLowerCase() === "false"
```
mbostock, line 566:
```suggestion
        const trimValue = value.trim().toLowerCase();
        return trimValue === "true"
          ? true
          : trimValue === "false"
```

src/table.js
```diff
@@ -684,10 +762,102 @@ export function __table(source, operations) {
...
+      if (!typeCounts[col]) typeCounts[col] = initKey();
+      const type = typeof d[col];
+      const value = type === "string" ? d[col].trim() : d[col];
+      if (type !== "string") {
+        if (Array.isArray(value)) ++typeCounts[col].array;
+        else if (value instanceof Date) ++typeCounts[col].date;
+        else if (value instanceof ArrayBuffer) ++typeCounts[col].buffer;
+        else if (type === "number") {
+          ++typeCounts[col].number;
+          if (Number.isInteger(+value)) ++typeCounts[col].integer;
```
mbostock, line 832:
```suggestion
          if (Number.isInteger(value)) ++typeCounts[col].integer;
```

(Here we already know that `value` is a number so no coercion is necessary.)

src/table.js
```diff
@@ -684,10 +762,102 @@ export function __table(source, operations) {
...
+export function inferSchema(source, columns = getAllKeys(source)) {
+  const schema = [];
+  const sampleSize = 100;
+  let sample = source.slice(0, sampleSize);
+  const typeCounts = {};
+  for (const d of sample) {
+    for (const col of columns) {
+      if (!typeCounts[col]) typeCounts[col] = initKey();
+      const type = typeof d[col];
+      const value = type === "string" ? d[col].trim() : d[col];
```
mbostock, line 825:
```suggestion
      const value = d[col];
```

It would be cleaner to declare a trimmed value in the `else {` block below (to avoid another `type === "string"` test).

src/table.js
```diff
@@ -684,10 +762,102 @@ export function __table(source, operations) {
...
+        else if (value instanceof ArrayBuffer) ++typeCounts[col].buffer;
+        else if (type === "number") {
+          ++typeCounts[col].number;
+          if (Number.isInteger(+value)) ++typeCounts[col].integer;
+        }
+        // bigint, boolean, or object
+        else if (type in typeCounts[col]) ++typeCounts[col][type];
+        else if (value !== null && value !== undefined) ++typeCounts[col].other;
+      } else {
+        if (value.toLowerCase() === "true" || value.toLowerCase() === "false") {
```
mbostock, line 838: We could avoid calling toLowerCase() twice here either by having a local variable (similar to trimValue, say lowerValue) or by switching to a regular expression.

```suggestion
        if (/^(true|false)$/i.test(value)) {
```

src/table.js
```diff
@@ -684,10 +762,102 @@ export function __table(source, operations) {
...
+          if (Number.isInteger(+value)) ++typeCounts[col].integer;
+        } else if (/^\d+n$/.test(value)) ++typeCounts[col].bigint;
+        else if (DATE_TEST.test(value)) ++typeCounts[col].date;
+        else if (value) ++typeCounts[col].string;
+      }
+    }
+  }
+  for (const col in typeCounts) {
+    let type = greatest(Object.keys(typeCounts[col]), d => typeCounts[col][d]);
+    const numDefined = sample.filter(d => !(d[col] == null || d[col] === "")).length;
```
mbostock, line 852: This is testing for the empty string, but not the trimmed empty string i.e. the all-whitespace string. Hence this is inconsistent with the tests above. Maybe what we should do instead is count the number of non-missing values in the iteration above, so that we don’t have to repeat a similar test here for what is missing.

mbostock (changes requested):
I posted a few trivial suggestions which you can take or leave, but here is what I think (hope!) is the last logical issue! 🙏

src/table.js
libbey-observable: I really like the simplicity of creating and incrementing a `numDefined` property, but when I tried implementing it, it felt like it started to get complicated to figure out where/when to increment it. So I added a new check that takes the trimmed string into account... let me know what you think?

src/table.js
mbostock: I think we have to do this because string is special: it’s the “untyped” format we use to serialize other types (namely in CSV).

mbostock (approved):

PR #349, opened 2023-02-07
Rename type to types

mkfreeman:
This PR renames the key `.type` to `.types` on the incoming `operations` argument for compatibility with the data structure on Observablehq.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -618,8 +618,8 @@ export function __table(source, operations) {
   }
   // Combine column types from schema with user-selected types in operations
   const types = new Map(schema.map(({name, type}) => [name, type]));
-  if (operations.type) {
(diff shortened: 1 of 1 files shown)


src/table.js
```diff
@@ -618,8 +618,8 @@ export function __table(source, operations) {
   }
   // Combine column types from schema with user-selected types in operations
   const types = new Map(schema.map(({name, type}) => [name, type]));
-  if (operations.type) {
-    for (const {name, type} of operations.type) {
+  if (operations.types) {
+    for (const {name, type} of operations.types) {
```
annie, line 622: tiiiny nit: we should update the reference to `operations.type` in the comment on line 632

annie (approved):
👌

PR #352, opened 2023-02-17
duckdb-wasm to the latest version

kimmolinna:
#350 is to the previous version and this one is to the latest one.


mbostock:
#350 has been updated. Thanks.

PR #356, opened 2023-03-03
@duckdb/duckdb-wasm 1.24

mbostock:
Supersedes #350.

### DIFF
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -126,6 +126,9 @@ export class DuckDBClient {
     if (config.query?.castTimestampToDate === undefined) {
       config = {...config, query: {...config.query, castTimestampToDate: true}};
     }
+    if (config.query?.castBigIntToDouble === undefined) {
+      config = {...config, query: {...config.query, castBigIntToDouble: true}};
+    }
     await db.open(config);
     await Promise.all(
       Object.entries(sources).map(async ([name, source]) => {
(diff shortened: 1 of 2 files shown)


src/duckdb.js
```diff
@@ -169,7 +171,7 @@ async function insertFile(database, name, file, options) {
     const buffer = await file.arrayBuffer();
     await database.registerFileBuffer(file.name, new Uint8Array(buffer));
   } else {
-    await database.registerFileURL(file.name, url, duckdb.DuckDBDataProtocol.HTTP);
+    await database.registerFileURL(file.name, url, 4); // duckdb.DuckDBDataProtocol.HTTP
```
domoritz, line 174: This could be an issue if the enum ever changes.

mbostock: Yes. Something to check when we upgrade.

domoritz: Is there a reason not to use the value from the DuckDB lib?

mbostock: Just plumbing. The `duckdb` module isn’t available globally; it’s loaded in `loadDuckDB` here:

https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/duckdb.js#L256

So, I’d need to call `loadDuckDB` again here, or I’d need to somehow pass it through the call to `createDuckDB` above and then pass it down into `insertFile`.

https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/duckdb.js#L127

https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/duckdb.js#L138

All of those felt like a lot more work than typing the number `4`.

PR #357, opened 2023-03-08
Separate loadChartDataSource from loadTableDataSource

lileeyuh:


src/table.js
```diff
@@ -196,6 +196,22 @@ function sourceCache(loadSource) {
...
+      case "text/csv": return source.csv({typed: true});
+      case "text/tab-separated-values": return source.tsv({typed: true});
+      case "application/json": return source.json();
+    }
+   
+    throw new Error(`unsupported file type: ${source.mimeType}`);
+  }
+
+  if (isDataArray(source) && arrayIsPrimitive(source))
+    return Array.from(source, (value) => ({value}));
```
mbostock, line 211: I believe we disallow array-of-primitives as a data source type for chart cells, so this shouldn’t be necessary (and the compiler doesn’t support it anyway).

```suggestion
```

src/table.js
```diff
@@ -196,6 +196,22 @@ function sourceCache(loadSource) {
...
 }
 
+const loadChartDataSource = sourceCache(async (source) => {
+  if (source instanceof FileAttachment) {
+    switch (source.mimeType) {
+      case "text/csv": return source.csv({typed: true});
+      case "text/tab-separated-values": return source.tsv({typed: true});
+      case "application/json": return source.json();
+    }
+
```
mbostock, line 206: Style nit.

```suggestion
```

mbostock (changes requested):

src/table.js
lileeyuh: Thanks for catching!

mbostock (approved):

PR #358, opened 2023-03-08
Export inferSchema

annie:
Supports [this fix](https://github.com/observablehq/observablehq/pull/10916), which makes schema inference consistent across stdlib and the monorepo.


mkfreeman (approved):
LGTM!

PR #360, opened 2023-03-18
Add file.csv({typed: "auto"}) option that uses inferSchema

tophtucker:
Since 2020, file.csv and file.tsv have allowed you to pass {typed: true} to apply d3.autoType, which looks line by line and tries to guess the types. More recently, the data table cell has introduced an internal [inferSchema function](https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/table.js#L835-L885), which instead looks at a sample of rows and picks the most frequent type.

This introduces a new {typed: "auto"} option, which calls inferSchema and then coerces all rows to that schema, allowing FileAttachment calls to match the behavior of the data table cell. It also uses that new "auto" option in loadChartDataSource.

This duplicates a couple lines of the __table function…

https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/table.js#L621

https://github.com/observablehq/stdlib/blob/89f27007453ea1919493dca3fb94b7a5bde03222/src/table.js#L630

…but, since that has some other logic for overriding types, I didn’t bother refactoring anything for now.

Note that, since the existing code only checks for the truthiness of typed, and “auto” is truthy, this could theoretically change the behavior of existing code, if anyone has somehow already said `typed: "auto"`. But I think that’d be very rare!!

### DIFF
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -2,8 +2,8 @@ import {autoType, csvParse, csvParseRows, tsvParse, tsvParseRows} from "d3-dsv";
 import {arrow4, arrow9, arrow11, jszip, exceljs} from "./dependencies.js";
 import {cdn, requireDefault} from "./require.js";
 import {SQLiteDatabaseClient} from "./sqlite.js";
-import {Workbook} from "./xlsx.js";
 import {inferSchema, coerceRow} from "./table.js";
+import {Workbook} from "./xlsx.js";
 
 async function remote_fetch(file) {
   const response = await fetch(await file.url());
@@ -11,18 +11,19 @@ async function remote_fetch(file) {
   return response;
 }
 
+function enforceSchema(source, schema = inferSchema(source)) {
+  const types = new Map(schema.map(({name, type}) => [name, type]));
+  return Object.assign(source.map(d => coerceRow(d, types, schema)), {schema});
+}
+
 async function dsv(file, delimiter, {array = false, typed = false} = {}) {
   const text = await file.text();
   const parse = (delimiter === "\t"
     ? (array ? tsvParseRows : tsvParse)
     : (array ? csvParseRows : csvParse));
-  if (typed === "auto") {
-    const source = parse(text);
-    const schema = inferSchema(source);
-    const types = new Map(schema.map(({name, type}) => [name, type]));
-    return source.map(d => coerceRow(d, types, schema));
-  }
-  return parse(text, typed && autoType);
+  return typed === "auto" 
+    ? enforceSchema(parse(text)) 
+    : parse(text, typed && autoType);
 }
 
 export class AbstractFile {
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -11,7 +11,7 @@ async function remote_fetch(file) {
   return response;
 }
 
-function enforceSchema(source, schema = inferSchema(source)) {
+export function enforceSchema(source, schema = inferSchema(source)) {
   const types = new Map(schema.map(({name, type}) => [name, type]));
   return Object.assign(source.map(d => coerceRow(d, types, schema)), {schema});
 }
--- a/src/fileAttachment.js
+++ b/src/fileAttachment.js
@@ -3,6 +3,7 @@ import {arrow4, arrow9, arrow11, jszip, exceljs} from "./dependencies.js";
 import {cdn, requireDefault} from "./require.js";
 import {SQLiteDatabaseClient} from "./sqlite.js";
 import {Workbook} from "./xlsx.js";
+import {inferSchema, coerceRow} from "./table.js";
 
 async function remote_fetch(file) {
   const response = await fetch(await file.url());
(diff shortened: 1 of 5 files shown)


src/fileAttachment.js
```diff
@@ -12,9 +13,16 @@ async function remote_fetch(file) {
...
   const text = await file.text();
-  return (delimiter === "\t"
-      ? (array ? tsvParseRows : tsvParse)
-      : (array ? csvParseRows : csvParse))(text, typed && autoType);
+  const parse = (delimiter === "\t"
+    ? (array ? tsvParseRows : tsvParse)
+    : (array ? csvParseRows : csvParse));
+  if (typed === "auto") {
+    const source = parse(text);
+    const schema = inferSchema(source);
```
mbostock, line 21: This is missing one of the things we talked about:

```suggestion
    const schema = inferSchema(source, source.columns);
```

We already know the columns thanks to dsvParse, so we don’t have to scan them again.

Even better, you could change `inferSchema` itself so that it peeks at source.columns rather than always defaulting to getAllKeys in the default expression for `columns`. That might be nice because as it is today, everyone who calls `inferSchema` has to remember to pass in `source.columns` if it exists.

src/fileAttachment.js
```diff
@@ -12,9 +13,16 @@ async function remote_fetch(file) {
 
 async function dsv(file, delimiter, {array = false, typed = false} = {}) {
   const text = await file.text();
-  return (delimiter === "\t"
-      ? (array ? tsvParseRows : tsvParse)
-      : (array ? csvParseRows : csvParse))(text, typed && autoType);
+  const parse = (delimiter === "\t"
+    ? (array ? tsvParseRows : tsvParse)
+    : (array ? csvParseRows : csvParse));
+  if (typed === "auto") {
```
mbostock, line 19:
```suggestion
  if (typed === "auto" && !array) {
```

coerceRow isn’t compatible with `array`; it’ll generate an object for each row, but we’re supposed to return an array-of-arrays. I guess we could change coerceRow to support generating arrays instead of objects, but I’m not sure it really makes sense to run inferSchema on the array case, because the array case is generally for when the CSV file has an unusual shape (_e.g._, a variable number of fields per line).

src/fileAttachment.js
```diff
@@ -12,9 +13,16 @@ async function remote_fetch(file) {
...
+    const source = parse(text);
+    const schema = inferSchema(source);
+    const types = new Map(schema.map(({name, type}) => [name, type]));
+    return source.map(d => coerceRow(d, types, schema));
```
mbostock, line 23: Calling `source.map` doesn’t propagate the `source.columns` field. We could expose that here, too, but even better would be to expose the `source.schema` that we’ve already computed so that it doesn’t have to be inferred again. (Per the DatabaseClient spec, `source.schema` takes precedence over `source.columns`, so we don’t need to expose both.)

Also, it’d be nice to have a function for this operation, like say

```js
function coerceRows(source, schema) {
    const types = new Map(schema.map(({name, type}) => [name, type]));
    return source.map(d => coerceRow(d, types, schema)); // TODO expose schema
}
```

or even folding in the schema inference if necessary…

```js
function enforceSchema(source, schema = inferSchema(source)) {
    const types = new Map(schema.map(({name, type}) => [name, type]));
    return source.map(d => coerceRow(d, types, schema)); // TODO expose schema
}
```

Doesn’t really matter too much though, since it’ll presumably be only called in one place. It just might be nice since you could write nice standalone tests for it. Feels like a well-defined, self-contained operation, I mean.

src/fileAttachment.js
```diff
@@ -3,6 +3,7 @@ import {arrow4, arrow9, arrow11, jszip, exceljs} from "./dependencies.js";
 import {cdn, requireDefault} from "./require.js";
 import {SQLiteDatabaseClient} from "./sqlite.js";
 import {Workbook} from "./xlsx.js";
+import {inferSchema, coerceRow} from "./table.js";
```
mbostock, line 6: Nit: sort imports by path? (VS Code’s “Organize Imports” command can do this for you automatically.)

mbostock (changes requested):
Pretty good but a few suggestions!

src/fileAttachment.js
tophtucker: Changed inferSchema to default columns to `source.columns || getAllKeys(source)`

src/fileAttachment.js
tophtucker: Nice 👍  Feels more readable with enforceSchema pulled out : ) and assigned the schema property to the returned array

tophtucker:
Thanks for the comments Mike! Mostly addressed I think, 

- [ ] Add test of default columns to inferSchema tests
- [ ] Add test of enforceSchema
- [ ] Don't explicitly pass source.columns in __table, now that it defaults to it? (Not sure about this)

mbostock (changes requested):
Sorry, but I realize we’re missing some validation here, and therefore I rescind my prior suggestion on defaults. (The defaults were just for our own convenience but make it harder to implement the correct behavior.)

mbostock (approved):
👍👍

tophtucker:
😅 🙏 😌 

Reverted adding the defaults! And it now fixes https://github.com/observablehq/observablehq/issues/11103:

image

PR #363, opened 2023-04-14
Add and export getSchema function

annie:
This PR adds a `getSchema` function, building on a [suggestion from Mike](https://github.com/observablehq/observablehq/pull/11496#discussion_r1165998001). This function conditionally infers a schema depending on whether or not a valid one exists on the source, combines the schema with any saved type assertions, and determines whether or not the source data needs to be coerced.

We will use `getSchema` in both `__table` and in the table schema tasks in the worker (I'll update https://github.com/observablehq/observablehq/pull/11496 to use this function), and thereby unify the two paths that we currently have for type inference.

I also added a missing unit test for type assertion in `__table`, as well as some tests for `getSchema`.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -636,8 +636,8 @@ export function __table(source, operations) {
   const schemaInfo = getSchema(source);
   let {columns} = source;
   let {schema, shouldCoerce} = schemaInfo;
-  const types = new Map(schema.map(({name, type}) => [name, type]));
   // Combine column types from schema with user-selected types in operations
+  const types = new Map(schema.map(({name, type}) => [name, type]));
   if (operations.types) {
     for (const {name, type} of operations.types) {
       types.set(name, type);
--- a/src/table.js
+++ b/src/table.js
@@ -622,9 +622,9 @@ export function getSchema(source) {
   let {schema} = source;
   if (!isQueryResultSetSchema(schema)) {
     schema = inferSchema(source, isQueryResultSetColumns(columns) ? columns : undefined);
-    return {schema, shouldCoerce: true};
+    return {schema, inferred: true};
   }
-  return {schema, shouldCoerce: false};
+  return {schema, inferred: false};
 }
 
 // This function applies table cell operations to an in-memory table (array of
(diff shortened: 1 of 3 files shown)


annie:
hm on second thought, maybe we should handle type assertions outside of `getSchema`, because it's a pain to have to pass around `operations.types` in the worker.

src/table.js
```diff
@@ -617,34 +617,38 @@ export function coerceToType(value, type) {
...
 }
 
+export function getSchema(source) {
+  const {columns} = source;
+  let {schema} = source;
+  if (!isQueryResultSetSchema(schema)) {
+    schema = inferSchema(source, isQueryResultSetColumns(columns) ? columns : undefined);
+    return {schema, shouldCoerce: true};
+  }
+  return {schema, shouldCoerce: false};
```
mbostock, line 627: Maybe we should call this “inferred” instead of “shouldCoerce”? Because “inferred” makes a factual observation about how the schema was derived, whereas “shouldCoerce” implies that the caller should do something with the returned schema—and perhaps the caller wants to do something else?

annie: good call, updated.

src/table.js
```diff
@@ -617,18 +617,25 @@ export function coerceToType(value, type) {
...
   const input = source;
-  let {schema, columns} = source;
-  let inferredSchema = false;
-  if (!isQueryResultSetSchema(schema)) {
-    schema = inferSchema(source, isQueryResultSetColumns(columns) ? columns : undefined);
-    inferredSchema = true;
-  }
+  const schemaInfo = getSchema(source);
+  let {columns} = source;
+  let {schema, inferred} = schemaInfo;
```
mbostock, line 638: Minor simplification:

```suggestion
  let {columns} = source;
  let {schema, inferred} = getSchema(source);
```

I don’t think we should fix it now, but there’s a latent bug below where where we’re assuming that if `source.columns` is truthy, it’s a valid QueryResultSetColumns. That’s because this function is trying to preserve the same shape of output as the input; so, if the input had a source.columns, we return a result.columns. I don’t think we _need_ to do that; we could just always return a result.schema instead, now that we’re inferring a schema as needed. But again, reasonable to do that as a separate change to minimize risk since it’s not directly related to this PR’s objective.

mbostock (approved):

src/table.js
annie: hm yeah, now that `.schema` is always defined, do we even need to return `.columns` from this function or can we just remove it? or would that be a breaking change, since we export `__table` from stdlib?

mbostock: Right, I meant I think that __table can always return .schema (and never .columns). I don’t _think_ that would be breaking… 🤷

PR #365, opened 2023-04-20
Remove .columns

annie:
(Based on a suggestion [from Mike](https://github.com/observablehq/stdlib/pull/363#discussion_r1167258471))

Remove .columns from `__table`. It's no longer needed since we always guarantee that we'll return a .schema, and .schema is a superset of the information in .columns.

One thing I'm not too sure about – if the source has a .columns property and the source isn't modified in `__table`, the .columns property will pass through and remain in the output. Is that ok, or do we need to standardize `__table` by making sure .schema is the only property on the output?

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -633,7 +633,6 @@ export function getSchema(source) {
 // function to do table operations on in-memory data?
 export function __table(source, operations) {
   const input = source;
-  let {columns} = source;
   let {schema, inferred} = getSchema(source);
   // Combine column types from schema with user-selected types in operations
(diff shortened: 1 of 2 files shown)


libbey-observable (approved):
These changes look good to me. If a source comes in to `__table()` with .columns and thus passes through and gets returned with .columns, that seems okay to me, but if we want to dig in deeper on that I'd be up for that, too.

annie:
yep that makes sense to me! thinking about it a bit more, there could be any number of properties on `source` that we just pass through right now, so it makes sense to treat `.columns` as just one of those.

PR #367, opened 2023-04-27
Derived columns

annie:
Part of https://github.com/observablehq/observablehq/issues/11623

## Description

This PR adds support for derived columns in the `__table` function. In this first version, we won't support derived columns for database sources.

Notable changes:
- Pulled the logic for inferring types and coercing rows into a separate `applyTypes` function. We now run this twice, first on the source dataset, and second on the derived dataset, before merging the two. Type inference/coercion must be done separately on the derived dataset because it may depend on values in the source dataset already being coerced.
- Added a `.fullSchema` property to the return value of `__table`. This property contains the schema information for _all_ columns in the dataset, regardless of whether or not they are selected (`.schema` only contains the schema info for selected columns). In https://github.com/observablehq/observablehq/pull/11214, we switch to using the cell value to get the table schema, because derived columns aren't available on the original dataset and their types may be dynamic, so we need to look at their evaluated runtime values. We look at `.fullSchema` when fetching the table schema so that we always have type information for the full set of columns, so that users can e.g. reselect a deselected column in the Columns menu.

## Review notes

I would love some feedback on the `.fullSchema` change! I'm not sure if it's the best way to make all the column types available, and I'm also not sure if I'm missing any major pitfalls with switching to use the cell value to fetch the table schema, instead of looking at the original data source as we do today. The main pitfall I experienced when testing is that, if there's an error in a derived formula, we no longer have a table schema available because the cell throws an error. I addressed that in the monorepo PR by adding a fallback that goes back to using the loaded data source + an approximation of the derived columns schema, which I think is the best we can do in that case. But perhaps there are other issues with this approach that I'm missing...

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -156,6 +156,7 @@ export const __query = Object.assign(
     if (isDatabaseClient(source)) return evaluateQuery(source, makeQueryTemplate(operations, source), invalidation);
     if (isDataArray(source)) {
       if(operations.derive) {
+        source = applyTypes(source, operations);
         operations.derive.map(({name, value}) => {
           source = source.map((row, index, rows) => ({...row, [name]: value(row, index, rows)}));
         });
@@ -612,11 +613,10 @@ export function coerceToType(value, type) {
   }
 }
 
-// This function applies table cell operations to an in-memory table (array of
-// objects); it should be equivalent to the corresponding SQL query. TODO Use
-// DuckDBClient for data arrays, too, and then we wouldn’t need our own __table
-// function to do table operations on in-memory data?
-export function __table(source, operations) {
+// This function uses schema information and user-asserted types to coerce table
+// values. This is a prerequisite to being able to derive columns.
+// TODO: Can we add a `typed` flag somewhere to prevent doing this twice?
+function applyTypes(source, operations) {
   const input = source;
   let {schema, columns} = source;
   let inferredSchema = false;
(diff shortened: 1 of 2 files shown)


src/table.js
```diff
@@ -651,6 +649,55 @@ export function __table(source, operations) {
...
+      // compute derivations on the renamed source.
+      // TODO Allow derived columns to reference other derived columns.
+      applyNames(source, operations).map((row, index, rows) => {
+        const resolved = value(row, index, rows);
+        if (derivedSource[index]) {
+          derivedSource[index] = {...derivedSource[index], [name]: resolved};
+        } else {
+          derivedSource.push({[name]: resolved});
+        }
+      });
```
libbey-observable, line 692: It's possible this is my misunderstanding, but based on this bit here, I expected my derived column values to remain intact after re-naming the column they were based on. Quick demo of what happened instead:

https://github.com/observablehq/stdlib/assets/111310561/9f2693db-ce5f-495f-8eb6-09ebe97aea77

annie: oooh that's a great catch!! unfortunately this bit of code doesn't support that case – it makes sure that the column derivations run on the renamed set of columns, but doesn't make sure that the column names in the formula are up-to-date. to make that work, we'd have to modify the formula to replace all instances of the old name with the new name at the time of renaming the column. i can add that to the list of things to fix (it'll require a monorepo change, so shouldn't affect stdlib stuff).

libbey-observable: Oh, of course! I see that now. Thanks.

annie:
just pushed up a new commit that catches runtime errors and returns them in the function output (although it seems like i need to update tests!). will put up a corresponding monorepo PR shortly!

libbey-observable (approved):

PR #373, opened 2023-06-07
Catch more invalid data sources

annie:
Companion PR to https://github.com/observablehq/observablehq/pull/12427

If a user selects a cell with a primitive value, the call to `cache.get` throws an "invalid value used as WeakMap key" error. I think it's more informative to show "invalid data source" in that case, and in general I'd like to standardize the error message across all cases where the data source isn't valid.

### DIFF
--- a/src/table.js
+++ b/src/table.js
@@ -184,7 +184,7 @@ export async function loadDataSource(source, mode, name) {
 function sourceCache(loadSource) {
   const cache = new WeakMap();
   return (source, name) => {
-    if (!source) throw new Error("data source not found");
+    if (!source || !(source instanceof Object)) throw new Error("invalid data source");
     let promise = cache.get(source);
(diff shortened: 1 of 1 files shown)


src/table.js
```diff
@@ -184,7 +184,7 @@ export async function loadDataSource(source, mode, name) {
 function sourceCache(loadSource) {
   const cache = new WeakMap();
   return (source, name) => {
-    if (!source) throw new Error("data source not found");
+    if (!source || !(source instanceof Object)) throw new Error("invalid data source");
```
mbostock, line 187:
```suggestion
    if (!source || typeof source !== "object") throw new Error("invalid data source");
```

mbostock (changes requested):

mbostock (approved):

PR #375, opened 2023-06-16
5.7.6

tophtucker:
After running resolve-dependencies, I omitted the two major upgrades: arquero 4.8.8 → 5.2.0 and mermaid 9.2.2 → 10.2.3 (though latter would prob be fine?).

Should we test duckdb 1.27.0 a little? I know we've been sensitive to those upgrades in the past.


mbostock:
You should test everything you’re upgrading. So, unless you also want to test Vega & DuckDB, you might want to omit those for now.

mbostock (approved):

tophtucker:
tested changes fyi https://observablehq.com/d/4303b3734fb950d8

PR #382, opened 2023-08-09
5.8.0

annie:
i used to bump the version as part of `npm publish`, but it looks like i can't push directly to main anymore (was there a change in a the repo settings recently?).

anyway, this applies the version bump that i neglected to add to https://github.com/observablehq/stdlib/pull/381.


tophtucker (approved):
Yeehaw!

PR #389, opened 2023-11-18
absolute DuckDB FileAttachment URL

Fil:
I've only tested this locally (not sure how unit tests could work).

### DIFF
--- a/src/duckdb.js
+++ b/src/duckdb.js
@@ -208,7 +208,7 @@ async function insertFile(database, name, file, options) {
         }
         if (/\.parquet$/i.test(file.name)) {
           return await connection.query(
-            `CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`
+            `CREATE VIEW '${name}' AS FROM parquet_scan('${file.name}')`
           );
         }
         throw new Error(`unknown file type: ${file.mimeType}`);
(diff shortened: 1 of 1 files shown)


src/duckdb.js
```diff
@@ -171,7 +171,7 @@ async function insertFile(database, name, file, options) {
     const buffer = await file.arrayBuffer();
     await database.registerFileBuffer(file.name, new Uint8Array(buffer));
   } else {
-    await database.registerFileURL(file.name, url, 4); // duckdb.DuckDBDataProtocol.HTTP
+    await database.registerFileURL(file.name, new URL(url, document.location).href, 4); // duckdb.DuckDBDataProtocol.HTTP
```
mbostock, line 174:
```suggestion
    await database.registerFileURL(file.name, new URL(url, location).href, 4); // duckdb.DuckDBDataProtocol.HTTP
```

src/duckdb.js
```diff
@@ -208,7 +208,7 @@ async function insertFile(database, name, file, options) {
         }
         if (/\.parquet$/i.test(file.name)) {
           return await connection.query(
-            `CREATE VIEW '${name}' AS SELECT * FROM parquet_scan('${file.name}')`
+            `CREATE VIEW '${name}' AS FROM parquet_scan('${file.name}')`
```
mbostock, line 211: Is the shorter form recommended? Seems like the docs still use `SELECT *`.

https://duckdb.org/docs/data/parquet/overview.html

mbostock (approved):

src/duckdb.js
Fil: It's brand new. Documented in https://duckdb.org/docs/sql/query_syntax/from with one sentence “DuckDB also has an optional FROM-first syntax which enables you to also query without a SELECT statement.” The [announcement](https://duckdb.org/2023/08/23/even-friendlier-sql.html#from-first-in-select-statements) (August 2023) mentions that `SELECT *` is "obsolete".

Fil: I'll revert that commit for now, since I'm not sure when the new syntax was introduced and it's not related to the FileAttachment issue.

mbostock (approved):
Re-approving. 😁

Fil:
I feel compelled to click on the ✅

PR #399, opened 2026-01-02
Reject pending promises when a generator is closed

tomlarkworthy:
This treats an issue raised here https://github.com/observablehq/runtime/issues/376

V6 runtime will wait for prior promises to resolve, in the case of a generator like Generator.input, the initial value will wait for a truthy value, if the cell is invalidated, the initial value will never be resolved and the runtime is stuck.

In this change we reject any pending promises when a Generator is disposed, ensuring nothing gets hung. By rejecting we ensure nothing interprets this as an actual value.

I can confirm this fixes a behaviour I was observing when running Notebook 1.0 on Runtime v6.0

### DIFF
--- a/src/generators/observe.js
+++ b/src/generators/observe.js
@@ -4,6 +4,7 @@ export function observe(initialize) {
   let stale = false;
   let value;
   let resolve;
+  let reject;
   const dispose = initialize(change);
 
   if (dispose != null && typeof dispose !== "function") {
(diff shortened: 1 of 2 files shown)


mbostock:
Do we need a similar fix here? Do you mind testing the new standard lib too?

https://github.com/observablehq/notebook-kit/blob/main/src/runtime/stdlib/generators/observe.ts

tomlarkworthy:
yes the issues exists in Notebook 2.0 too, bit harder to fix with async generator sugar so I converted out. 
https://github.com/observablehq/notebook-kit/pull/115
That PR against `notebook-kit` is not meant to be merged yet, more of a conversation tool but I can can tidy it up if you want.

tomlarkworthy:
I guess this one is not needed because the stdlib is intended for Runtime v5 which doesn't have the same contract on generators.

